PR #13, opened 2012-12-07
I added a cross-platform implementation of pyodbc.drivers()

sergiodlc:
Hi Michael:

I removed the windows only version of pyodbc.drivers() and added an OS independent version of it, along with its tests. I tested that with iODBC in Ubuntu 12.04 x86.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -155,8 +155,20 @@ def get_compiler_settings(version_str):
 
         settings['define_macros'].append( ('MAC_OS_X_VERSION_10_7',) )
 
+    elif sys.platform.startswith('linux'):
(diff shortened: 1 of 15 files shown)


apatil:
Hi @mkleehammer, is this PR still in the pipeline?

mkleehammer:
Merged the drivers implementation.  Sorry it took so long - I thought this was a different patch.

I did not merge the 2nd commit - I'd need some more time with it, especially with systems that have both installed for testing, etc.  Plus I think I'd like a warning if it picks iODBC which seems to be falling out of favor and may have some SQLWCHAR issues.

PR #20, opened 2013-10-12
Add SQL_WCHART_CONVERT for UnixODBC. 

maparent:
The simple test in testunicode.py fails otherwise, on Ubuntu 64 i686, with a 64 bit driver. (Virtuoso)
This is not yet final code, I will package it better once I hear confirmation that it does not break other configurations, or create tests otherwise.
I'm sending it as a pull request in this state to raise the issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -161,6 +161,9 @@ def get_compiler_settings(version_str):
         # Python functions take a lot of 'char *' that really should be const.  gcc complains about this *a lot*
         settings['extra_compile_args'] = ['-Wno-write-strings']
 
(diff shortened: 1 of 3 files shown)


mkleehammer:
Now that I look at this again I'm not sure it is a good idea.  The unixODBC authors discourage changing flags like using wchar_t size, etc.  Can you try 3.0.10 and see if it helps.

A better solution might be to allow an encoding to be set on a connection such as UTF8.  I also might be able to optimize it a bit in Python 3.4+ where we might actually have 1-byte strings internally.

mkleehammer:
I have rewritten Unicode support in the v4 branch.

PR #23, opened 2014-03-06
Fixes #22 - Increase number of loop iterations to fully pull in data

### LINKED ISSUE #22: Not returning all data
Linking this issue to git.
https://code.google.com/p/pyodbc/issues/detail?id=125

geota:
Increase number of loop iterations to fully pull in data from VARCHAR 65000 fields.

The current pyodbc behavior will never pull in more than 1024 + (2048 * 9) = 19456 bytes.


JeffAMcGee:
I was pulling data from a Vertica database, and if a field contained more than 9728 characters, it would return nothing. This patch worked for me. Thanks.

PR #32, opened 2015-01-13
Update setup.py

diogommartins:
Fixes problems related with "ld: library not found for -lodbc" during the build phase. Considering you're compiling on OS X 10.9 or 10.10 and installed unixodbc with homebrew

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -160,13 +160,14 @@ def get_compiler_settings(version_str):
 
     elif sys.platform == 'darwin':
         # OS/X now ships with iODBC.
-        settings['libraries'].append('iodbc')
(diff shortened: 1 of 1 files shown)


diogommartins:
Yep. I'm using 10.10 also and the compiler had some trouble without the extra_link_args

mkleehammer:
I'm not seeing that, but I am on the latest 10.10.4 beta and my Xcode is out of date.  I should grab the latest to match the beta.

What versions do you have?

The installation method of unixODBC could make a difference too.  I installed via Homebrew, I think, which may have put the libs somewhere Xcode would pick them up automatically?  I'll have to look into this.  How did you install unixODBC?

PR #33, opened 2015-02-03
Trapping errors on nextset (plus couple of bugfixes which prevented compilation)

tjguk:
As things stood, a nextset would error out with an unhelpful ("HY000") error if it encountered a SQL exception or an explicit RAISERROR. This patch builds the exception before freeing the cursor.

While building, there was one compilcation error on and one runtime error which I've also patched. (I'm running on Windows 7, Python 2.7-32 against SQL 2008 but I've tested on Python 3.4-64 as well).


mkleehammer:
Unfortunately I missed the first two important fixes and had to track them down myself.  Because of that I simply cherry picked the nextset fix and will close the PR.  Thanks for the fix.

PR #66, opened 2015-07-15
Add interval support

tekknolagi:
* Make SQL_INTERVAL_* types of pytype PyDateTimeAPI->DeltaType
* Add date string parser
* Add GetDataInterval handler and associated case in GetData

### DIFF
--- a/src/cursor.cpp
+++ b/src/cursor.cpp
@@ -188,6 +188,22 @@ static PyObject* PythonTypeFromSqlType(Cursor* cur, const SQLCHAR* name, SQLSMAL
         pytype = (PyObject*)PyDateTimeAPI->TimeType;
         break;
 
+    case SQL_INTERVAL_YEAR:
+    case SQL_INTERVAL_MONTH:
+    case SQL_INTERVAL_DAY:
+    case SQL_INTERVAL_HOUR:
+    case SQL_INTERVAL_MINUTE:
+    case SQL_INTERVAL_SECOND:
+    case SQL_INTERVAL_YEAR_TO_MONTH:
+    case SQL_INTERVAL_DAY_TO_HOUR:
+    case SQL_INTERVAL_DAY_TO_MINUTE:
+    case SQL_INTERVAL_DAY_TO_SECOND:
+    case SQL_INTERVAL_HOUR_TO_MINUTE:
+    case SQL_INTERVAL_HOUR_TO_SECOND:
+    case SQL_INTERVAL_MINUTE_TO_SECOND:
+        pytype = (PyObject*)PyDateTimeAPI->DeltaType;
+        break;
+
     case SQL_TYPE_TIMESTAMP:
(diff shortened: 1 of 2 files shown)


ostrowr:
hey @mkleehammer any updates on this? Thanks :)

mvanderh:
Hey I just realised I didn't have intervals on my Vertica DB and realised I needed this patch. Can you please merge? This is great

gabrielhaim:
@mkleehammer any updates on this? That would be extremely useful!

mkleehammer:
I'll need to make some tests for a database I have easy access to.  Do any of the other ODBC libraries return interval types, such as PostgreSQL?

tekknolagi:
@mkleehammer PostgreSQL has interval types: http://www.postgresql.org/docs/9.1/static/datatype-datetime.html
as does MySQL: http://dev.mysql.com/doc/refman/5.0/en/time.html

tekknolagi:
Hold on. Having a couple of git issues.

tekknolagi:
I might close this PR in favor of opening a clean one.

tekknolagi:
It looks like with the patch PostgreSQL returns a time interval string, and MySQL has a floating point exception on my system (both before and after patch).

mkleehammer:
Two discussions need to be had:

We can't use STL functions that raise exceptions without catching them and converting them.  I try to use a safe subset of C++ that isn't going to raise unexpected exceptions.  Otherwise we're going to get a crash instead of a clean exit with a Python exception.  Perhaps that is what is going on with MySQL.  Can this be done with more or less straight C?

Second, the documented ODBC interface for intervals is here: [Interval Data Types](https://msdn.microsoft.com/en-us/library/ms716506(v=vs.85).aspxhttps://msdn.microsoft.com/en-us/library/ms716506(v=vs.85).aspx).  We should be calling GetData with a SQL_INTERVAL_STRUCT and returning an error if a driver doesn't support it.

I'm not clear where the format that the `read_time_token_list` function is reading is coming from.  If it is just what one DB happens to read when asking for data as a string, it won't work across databases.

Thoughts?

tekknolagi:
Which STL functions are you referring to? And no, the MySQL crash happens even with a clean fork from your repository.

I think you are right regarding the string — we have developed this primarily to work with Vertica. Can we add support for databases even if the driver does not support it?

mkleehammer:
Almost all STL functions can throw exceptions, including all vector methods like push_back.  However, if you can get Vertica to return a SQL_INTERVAL_STRUCT (by asking for one in GetData) then that issue would go away anyway.  A dynamic container wouldn't be needed to parse.

mkleehammer:
Good news - #80 is simpler interval patch.  I do want to update it so it isn't available just yet.  I am, however, going to close this one.  Thank you very much for supplying something, but I think you'll like #80 just as much.

Issue #65, opened 2015-07-08
Add property to check if a connection is open.


dalee-bis:
Trying to perform an operation on a closed connection currently produces a ProgrammingError but there is no way to check if a connection is open.

It would be great to be able to check if a connection is open before trying to run queries or close the connection.

tekknolagi:
@dalee-bis Just added this :)

jaymegordo:
+1 for this!

PR #70, opened 2015-07-28
Add is_closed method to Connection class

tekknolagi:
This helps avoid ProgrammingErrors, as per https://github.com/mkleehammer/pyodbc/issues/65

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -270,6 +270,23 @@ PyObject* Connection_New(PyObject* pConnectString, bool fAutoCommit, bool fAnsi,
     return reinterpret_cast<PyObject*>(cnxn);
 }
 
+static char is_closed_doc[] =
+    "Check whether the connection is closed.";
+
+static PyObject* Connection_is_closed(PyObject* self, PyObject* args)
+{
+    UNUSED(args);
+
+    Connection* cnxn = (Connection*)self;
+    if (!cnxn)
+        return 0;
+
+    if (cnxn->hdbc == SQL_NULL_HANDLE)
+        Py_RETURN_TRUE;
+    else
+        Py_RETURN_FALSE;
+}
+
 static void _clear_conv(Connection* cnxn)
 {
     if (cnxn->conv_count != 0)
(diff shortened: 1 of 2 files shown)


tekknolagi:
I may have misunderstood how to add tests. I'm looking into it.

tekknolagi:
Actually, I'm not even sure how to add tests.

mvanderh:
Oh! This is peachy :)

mkleehammer:
How do you feel about a property `closed` instead?  Would that be a bit nicer to use?

keitherskine:
A `closed` property would be more in alignment with file objects, so yes I personally would prefer that.  However, what would "closed" actually mean?  If the database itself spontaneously closes the connection (for whatever reason), would this property reflect that?

mkleehammer:
That would be useful, but it isn't something ODBC supports.  You can always try "select 1" or something.  Some databases have small stored procedures you can use to check a connection too.  In this case, I just meant I would prefer a `cnxn.closed` property to `cnxn.is_closed()`.

keitherskine:
OK.  Either way though, I certainly agree that a `closed` property is better than a function call.

tekknolagi:
How would you propose doing a property?

v-chojas:
You may be able to use the `SQL_ATTR_CONNECTION_DEAD` connection attribute to determine liveness.

Some drivers, such as the ODBC Driver for SQL Server, have an additional `SQL_COPT_SS_CONNECTION_DEAD` connection attribute which may also work.

NewVadim:
Hello!
Please release this feature with any `is_closed` or `closed` methods. Almost three years have passed.

robfraz:
Any news on getting this into a new release?...

mkleehammer:
OK - I'll work on it.  The last we left it everyone wanted to convert it to a property but that never happened.  I try not to mess with people's patches so that their name is on the commit and they get credit, but I'll add a `closed` property.

tekknolagi:
@mkleehammer go for it. It looks like the simplest route using this patch -- if you want to -- is just to make a Python-level property that wraps it. If you want to abandon it, go ahead.

tekknolagi:
Could not figure out how to update this PR so filed #924

PR #73, opened 2015-08-24
Applied patch for UTF-16 encoding on 64 bit machines

devintjones:
Had a problem encoding utf16 strings from Netezza on Centos 6.4 64bit. 

Found the fix/issue here:

https://code.google.com/p/pyodbc/issues/attachmentText?id=78&aid=5030789027258594097&name=wchar-fix.diff&token=ABZ6GAdn5zYg0bHJULckOP3EJLUwwp2ZtA%3A1440428010841

### DIFF
--- a/src/getdata.cpp
+++ b/src/getdata.cpp
@@ -316,7 +316,8 @@ static PyObject* GetDataString(Cursor* cur, Py_ssize_t iCol)
     case SQL_WCHAR:
     case SQL_WVARCHAR:
     case SQL_WLONGVARCHAR:
-        nTargetType  = SQL_C_WCHAR;
+        //nTargetType  = SQL_C_WCHAR;
+	nTargetType = SQL_C_CHAR;
         break;
 
     default:


mkleehammer:
I realize it works for you, but it isn't clear what else this is going to break.  If I remember correctly, that is going to cause it to use UTF-8 (SQLCHAR) instead of of UTF-16 (SQLWCHAR).  If that's not the case, then the bug is that it is being converting incorrectly and that should be fixed instead.

Since each database appears to be doing this differently (instead of following the specifications), I think the most general approach is to provide a set of defaults, but then allow each ODBC SQL type (SQL_WCHAR, SQL_XML, etc.) to have either (1) an encoding set, such as "utf16" or (2) a custom function for conversion.

The primary problem I have is everyone demonstrates these with databases I don't have.

Thoughts?

mkleehammer:
I have written Unicode support in the v4 branch - available as pull request #160.

Issue #60, opened 2015-06-29
Add interval support


tekknolagi:
I plan to submit a patch to fix support for intervals, particularly with a Vertica backend. Is there anything in particular I should read before submitting my patch? Perhaps regarding code style, licenses, etc?

I made a Google Groups post [here](https://groups.google.com/forum/#!topic/pyodbc/u_cqTusS7kM) but I am not sure if that is monitored much.

mkleehammer:
Sorry, I have been so busy on so many projects that I haven't been monitoring the group much.  As for style, just try to match the rest of the code.  Thanks.

tekknolagi:
No problem. Since I do not see many helper functions in `src/getdata.cpp`, do you have a particular style in mind?

mkleehammer:
I'm not really sure what your looking for, so feel free to start the pull request.  I alway recommend rebasing to a single commit anyway, so if you feel changes are needed afterwards it will end up fine.

I was assuming we were talking about the generic ODBC interval data type and you creating a Getinterval function or something.  Did I get that right?

tekknolagi:
That is correct, and it will be a single-commit patch.

tekknolagi:
Somewhat related: what do you want get_version() to report when not on a branch? Right now, since `git branch` shows `* (no branch)`, it reports something like `3.0.13b1+(no`, which is a) poorly formatted and b) messing up the `TOSTRING` macro. I propose a change that instead uses the commit hash when not on a branch.

mkleehammer:
I thought I fixed that once by replacing it with no-branch, but a hash would be great. Thanks. 

My intention is to ways be on a branch when developing so I ignored the few times I was using hit bisect, etc

tekknolagi:
Noted.

tekknolagi:
https://github.com/uber/pyodbc

http://i.imgur.com/7drHiqr.gif

tekknolagi:
Yay!
https://github.com/mkleehammer/pyodbc/pull/66
https://github.com/mkleehammer/pyodbc/pull/67

tekknolagi:
@mkleehammer Thoughts?

mweaverh:
Here's another attempt: https://github.com/mkleehammer/pyodbc/pull/80

tekknolagi:
Interesting.

2015-09-25 16:48 GMT-04:00 Matthew Weaver <notifications@github.com>:

bkline:
[...]

PR #80, opened 2015-09-25
More general support for SQL intervals

mweaverh:
https://github.com/mkleehammer/pyodbc/issues/60

### DIFF
--- a/src/cursor.cpp
+++ b/src/cursor.cpp
@@ -458,11 +458,16 @@ static void closeimpl(Cursor* cur)
     {
         HSTMT hstmt = cur->hstmt;
         cur->hstmt = SQL_NULL_HANDLE;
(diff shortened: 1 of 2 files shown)


mkleehammer:
Sorry for the delay - I somehow lost a bunch of emails.

I like the idea and added something similar to pglib, but I'm concerned about months and years.  For example, a lot of people would expect that adding a month would put you at the same day the month after, not just 30 days.  Because of the issues around this I raised an error when months or years are involved.

What do you think about that?

mweaverh:
Sounds good. Failure is preferable to erroneous results. 
I recall thinking it wouldn't be hard to represent year-month intervals precisely, but it's been a while since I've been in this code.

mkleehammer:
FYI - there is another merge request for interval support that looks interesting.  I'm waiting on tests for it.

Issue #62, opened 2015-07-03
Request to use parameter arrays in executemany()


keitherskine:
I just wanted to create this issue as a placeholder for the implementation of parameter arrays in pyodbc.

In cases where users are attempting to bulk load data into a database using the executemany() function, parameter arrays allow the data to be "buffered up" so that it is transferred in batches to the database rather than one record at a time.  This makes orders-of-magnitude differences in the speed of the operation.

Not all databases accept parameter arrays, but Microsoft SQL Server does:
https://msdn.microsoft.com/en-us/library/ms713841(v=vs.85).aspx

Github user 'gbegen' has apparently already implemented this functionality from an earlier fork of pyodbc (some years ago now), and his updates can be seen here:
https://github.com/gbegen/pyodbc/commit/956201ed27853c6445a03745bb0b99c3333262b5

It would be great if those changes could be folded into the latest version of pyodbc.  I for one would be extremely grateful.

chrisguilbeau:
We ran into this problem while inserting thousands of ids into a table. Our workaround was to break the id list into groups of 250 (**see below) and send a bunch of insert statements at once joined with a semi-colon along with the list of ids as parameters.

**We noticed if we'd send more than 500 statements, the function would return without complaining, but missing several inserts.

Fixing insert many to work as the ceODBC library does would be a great help.

GraemeBrown:
Yes, please! There's a thread on this on the (now closed?) Google Code page (https://code.google.com/p/pyodbc/issues/detail?id=250).

I am using ceodbc for all of my insert and update statements, and pyODBC for my select statements. It would be nice to just use a single connection and cursor for everything!

woodlandhunter:
It doesn't look like that could be sensibly merged. It seems there are many changes not related to parameter arrays. 

Also, is it really needed to add an extra kwarg to the connect method? Can pyodbc not check support for parameters arrays and do the right thing automatically?

lejordet:
I found this via the Google Code issue mentioned above, and accidentally posted on the wrong repo:
This is still happening in 3.0.10 for the record; I have a query that is significantly faster with ceODBC (MERGE of about 400k rows); 

[image](https://cloud.githubusercontent.com/assets/223602/13566853/f3e82d78-e457-11e5-9e84-364606350521.png)

This seems to be the main reason for slow bulk queries in pyodbc - the query in question completes in 15 seconds on ceodbc (which only prepares once), and takes 5 minutes on pyodbc.

[...]

PR #83, opened 2015-10-28
Fixing issue #62

vegitron:
Based on the first comment on issue #62, I sorted out the merge conflict between gbegen's fork and the mkleehammer base.

In my test, 10k inserts into a 1024 column table go from ~25 seconds to ~7 with parameterarraybinding=True.   ceODBC is ~4.5 seconds, so there's still room for improvement, but this is a big step.


vegitron:
Need to close while I figure out a unicode param issue

PR #119, opened 2016-05-30
Change absolute path to relative in setup.py

### LINKED ISSUE #82: raise ValueError("path '%s' cannot be absolute" % pathname)
When installing via pip, this error appears:

ValueError: path '/Users/mkleehammer/dev/pyodbc/src/buffer.cpp' cannot be absolute

ddwo:
This should fix #82

I don't see any build instructions with required dependencies for pyodbc so I'm unable to completely verify the bug is fixed with this commit but I'm able to create the SOURCES.txt file with `python setup.py egg_info` and the offending absolute paths disappear.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 
 import sys, os, re, platform
-from os.path import exists, abspath, dirname, join, isdir
+from os.path import exists, abspath, dirname, join, isdir, relpath
 
 try:
     # Allow use of setuptools so eggs can be built.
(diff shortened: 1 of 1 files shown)


christianklemp-imt:
@mkleehammer would like to see this fixed

maxwedwards:
Also would appreciate this being merged

maxwedwards:
To pip install before PR is merged: `pip install git+https://github.com/ddwo/pyodbc.git@relpath#egg=pyodbc`

mlipovsky:
Some update with fix?

s-trooper:
any progress here?

mkleehammer:
A heads-up for anyone who is working on related things later: The setup machinery will use existing egg info, so you need to delete your old egg-info directory before this fix takes effect.

PR #129, opened 2016-07-09
Fix SOURCES.txt paths

skillian:
Supposed to address issue #82.  It worked for me!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ def main():
 
     settings = get_compiler_settings(version_str)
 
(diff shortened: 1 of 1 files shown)


SuaveFool:
Thanks @skillian I was encountering the issue identified in #82 and saw your change and was able to finally get this working! I'm not sure whether #119 is a better solution across platforms/python versions, but this works with Win 7, Python 3.5.2 compiler VS 12.0.

skillian:
No problem, @SuaveFool, but I hadn't seen @ddwo's PR at the time that I made this one, but I do think that @ddwo's is better.

PR #131, opened 2016-07-31
Support column of type DATATIMEOFFSET and ODBC Driver 11 for SQL Server

aoblong:


mkleehammer:
Thanks for looking at this.

However, I'm not sure that returning it as text is the best approach.  This is inherently another datetime format, just with more precision and always in UTC with the UTC offset available.

I would really like to make them available to the user as a Python `datetime` with the appropriate time zone object, but those aren't standardized.

Have you considered using Connection.add_output_converter to make something that converts the data to what you need?

aoblong:
I agree with you, definitely my mistake. I was not aware of the connection output converter.

mkleehammer:
Thanks again for your work on this.  I'll close this now that you have a solution.

PR #139, opened 2016-09-08
adding homedir brew as a possible path

sgivens0:
for certain OSX environments, its not possible to use /usr/local as the prefix for brew, and the standard recommended prefix for alternative installs is $HOMEDIR/homebrew

cc @ilovezfs and @cartazio for external feedback on this homebrew lore.

This would help fix up a lot of issues for folks on my team, and possibly other enterprisey python3 users on OSX. 

many thanks!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -173,7 +173,7 @@ def get_compiler_settings(version_str):
         settings['define_macros'].append( ('MAC_OS_X_VERSION_10_7',) )
 
         # Add directories for MacPorts and Homebrew.
-        dirs = ['/usr/local/include', '/opt/local/include']
+        dirs = ['/usr/local/include', '/opt/local/include','~/homebrew/include']
         settings['include_dirs'].extend(dir for dir in dirs if isdir(dir))
 
     else:


ilovezfs:
https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Installation.md#alternative-installs

You can certainly install anywhere. The only major issue with that is that many of the prebuilt binary bottles aren't relocatable, which means that a source build will be triggered instead of just downloading and pouring the bottle.

Ideally, you would find a way to allow /usr/local to be writable, since that will be a much more efficient and reliable experience for the users, unless you intend to build your own bottles and host them, or unless it's acceptable to have people spending time each doing the same source builds, and spending time troubleshooting failed builds.

mkleehammer:
@ilovezfs Are you suggesting this should not be merged?

I've been holding off because this was a generic homebrew issue which I believe is now solved.  Does anyone believe it is not solved by recent homebrew / macOS updates?

ilovezfs:
I think it would be best to use /usr/local. Can you say more about why that's not possible? There is a specific exception for /usr/local such that it's writable with System Integrity Protection enabled, by default the directory doesn't exist, and Apple doesn't use it for anything it installs.

Regarding using other paths: that has been possible for years. brew determines the prefix based on the location of bin/brew.

The problem is efficiency. From a resources point of view, it is extremely wasteful to have every user individually build everything non-relocatable from source. The main way to overcome that if /usr/local is absolutely out of the question is to bottle everything your users will use yourself for some other prefix like /Users/Shared/Homebrew that can be the same on every machine. Third-party bottling is quite possible. I believe Facebook currently uses it as part of the build system for their software called osquery.

A significantly worse option, but still better than having everyone waste time building from source themselves, is to provide the installations already completed, either in a standardized prefix like /Users/Shared/Homebrew (in which case, or in /Users/$username/Homebrew before giving the machines to the users or in some sort of update (whatever your procedures allow for). However, you'll still end up with users either having to never update, which will leave security issues unpatched, or have them waste time building updates from source, or come up with a way of scripting the source builds of updates for them.

You should really find a way to use /usr/local unless resources are irrelevant, either measured in terms of IT burden to handle non-time-wasting installations and updates, or in terms of users' time redundantly building things from source and struggling with troubleshooting build failures, at which point they usually end up on Homebrew's doorstep requesting support.

PR #142, opened 2016-09-23
Forbid copy constructor use which may ends in an extra decref on objects

dgsb:
The tests against postgresql with python 3 are ok

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -21,7 +21,7 @@
 static char connection_doc[] =
     "Connection objects manage connections to the database.\n"
     "\n"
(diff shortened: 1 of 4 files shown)


dgsb:
@mkleehammer what do you think of this patch ?

mkleehammer:
I should have merged this ages ago as I walked right into the use of the generated copy ctor.  I don't remember what the GCC issue is but it could have been that I didn't provide a body for the private functions.  I don't remember what platform / version it was.

Anyway, I had to manually redo this so I'm closing this.  Thanks.

PR #151, opened 2016-10-30
python3 bug fix

stonebig:


mkleehammer:
Thanks for this, but this is actually a Python 2 test.  The tests2 directory is for Python 2 and tests3 directory is for python3.

PR #156, opened 2016-11-28
Update README.md

meet-bhagdev:


mkleehammer:
Sorry I let this sit so long.  I've moved all docs to the Wiki and minimized even the readme for now so I don't think this is needed any longer.  Thanks though.

PR #160, opened 2016-12-05
Rewrite Unicode handling for correctness and flexibility.

mkleehammer:
Reading and writing text now uses Python codecs which can be configured via the
new cnxn.setencoding and cnxn.setdecoding.  This is more complicated than simply
setting `cnxn.encoding = 'utf-16le'` but I've seen lots of weird combinations
that need to be handled more precisely.

This adds:

* connect(..., encoding=None)
* Connection.setencoding
* Connection.setdecoding

I'm not terribly happy with the API and would love some suggestions.  Perhaps we
should configure a "factory" with all of the settings that should be applied to
new connections?

I've also copied the docs from a separate branch into /docs which GitHub can
publish now.  I'm not sure how I want to handle documentation - there is a lot
of good information in the Wiki and I'm tempted to have minimal docs and cleanup
the Wiki.


mkleehammer:
@Lexcon Regarding a "raw" encoding, I've been thinking of exactly that.  For Python 2.7 it would make sense to accept that and default to returning `str`.

However, in the Python 3 and 2.7 `unicode` cases, it is not possible to create a Unicode object without *some* encoding.  If one is not provided then the system encoding (often "latin1" in US, I think) is used.

Which version were you planning on using it for?

If Python 3, how would you see creating these?

Lexcon:
Mike,

I think it would be nice to have an agnostic pass-through option, in 
which case it should not be a string but a bytes object that the user 
can convert on Python level.

For instance, this could be useful when a string is encrypted on Python 
level and actually just a series of bytes in the database, regardless 
that it's really stored in a varchar field.

Zipping is a similar case: I sometimes have a table where 99% of the 
fields fit in field width x, but in rare cases I have to zip the string 
and just add a leading marker to indicate so. Some database systems make 
a point of going over a certain length (for instance Oracle tends to 
have these legacy max-2000 byte field lengths). In both Oracle and SQL 
Server, on Python 2.7 I can do that without a problem now, but in Py3x I 
can't.

Robert

mkleehammer:
Agreed.  I have a global "converter" that can be configured by SQL type, but a per-column one after an execute (or as part of it) would be useful.  The existing converters receive the raw bytes and return the value that gets put into the Row object.  That would override the encodings, though, and be a separate feature.  I'll add an Issue for this - I'd like to see it as part of 4.0 also.

mkleehammer:
@Lexcon I have added "raw" as an encoding for Python 2.7 `str` types.

There just isn't a good way to build a Python 3 Unicode object directly from bytes without some kind of decoding.  The default is UTF-8.

Lexcon:
Mike,

This sounds just fantastic, I'll look into it over the weekend.

Robert

Lexcon:
Wouldn't it make sense to have the default behavior, in absence of 
parameters, unmodified compared to earlier versions?
I think otherwise existing software might break on upgrading pyodbc:

_py27, sqlserver2012_:
connection = pyodbc.connect(connectionstring, autocommit=True)
result = connectionb.cursor().execute(b"select 'aï' as uk, 'b' as 
jp").fetchall()

bombs. I would say the parameter, when defaulting to 'none', would 
ideally default to existing behavior.

Robert

mkleehammer:
@Lexcon I've done a lot of testing and SQL Server is working out-of-the-box on Windows.  Are you using the Linux driver?

You also didn't say what "bombs" does exactly, so I'm taking a guess.  If your Python 2.7 source file has extended characters then you have to declare the file's encoding at the top or you'll get this error:

    SyntaxError: Non-ASCII character '\xc3' in file C:\dev\pyodbc\tests2\test.py on line 32, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details

When I set my encoding to `# -*- coding: utf-8 -*-` the example SQL works fine.  If I set it to "latin-1" then it fails with "UnicodeDecodeError: 'utf8' codec can't decode byte 0xef in position 1: unexpected end of data".  That kind of makes sense to me since the characters are 0x61ef, but I haven't quite grokked why it displays properly (emacs).  Anyway, what encoding do you have?  I'm pretty sure it is working properly but I would like to get a complete understanding before releasing this to the world and getting 100K complaints the first week.

mkleehammer:
Another interesting data point - the error printed to the console displays "a∩" when I use the latin-1 encoding.  (I think macOS is using UTF-8 as the terminal encoding.)

Lexcon:
The code below, run on windows in Py27 with windows drivers works in 
3.1.1 but not in 3.1.2b6. Note that the sample code does not contain 
'from __future__ import unicode_literals'

import r4a.pyodbc as pyodbc
print(pyodbc.version)
connectionstring = 'DRIVER=ODBC Driver 11 for SQL 
Server;SERVER=DOTDATA2015\SQLEXPRESS;UID=yyyyy;PWD=xxxxx;AutoTranslate=no'
connection = pyodbc.connect(connectionstring, autocommit=True)
statement = "select 'aï' as uk, 'b' as jp".encode('utf-8')
result = connection.cursor().execute(statement).fetchall()
print(result)_
_

_OUTPUT on pyodbc 3.1.1:_

3.1.1
[('a\xc3\xaf', 'b')]

OUTPUT on pyodbc 3.1.2b6:
3.1.2b6
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 9: 
ordinal not in range(128)

Robert

mkleehammer:
@Lexcon There is a difference between the two code samples that makes the difference.  Try removing `.encode('utf-8')` from the select statement.

What's happening is that code changes the bytes from your internal encoding to UTF-8.  Internally *those* bytes are converted again using the UTF-8 algorithm and it corrupts it.

Lexcon:
Without the .encode('utf-8') part it definetely does not work in my 
3.1.2b6 version but it does in my 3.1.1 version.

My original code had no .encode('utf-8') part until I started tweaking 
it, this led to sending me the wrong code.

It's not a big issue to me if it's a bother to you, I just wanted to 
inform you about the incompatibility.

Robert

mkleehammer:
@Lexcon I see.  It's not a bother - I would like to get to the bottom of it.  At this point I think it is the encoding of your file.  I'm honestly not sure how it is working if you don't have the encoding set for your Python *source file*.  That is, how is what you are seeing on the screen encoded in the .py file that the interpreter reads.

If I use your example, I get the following error before the code even runs.  The Python 2.7.10 interpreter (default macOS version) cannot read the file:

    SyntaxError: Non-ASCII character '\xc3' in file tests2/test.py on line 34, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details

According to [that PEP](https://www.python.org/dev/peps/pep-0263) I think the right answer is that the source file encoding needs to be set to something compatible.

BTW, I can't recommend Python 3 enough.  I'm using 2.7 on a large project that we have not been able to convert, but all v3 stuff I do has been much more enjoyable, particularly when Unicode is involved.

I'm pretty sure I want to keep the defaults as-is, but I am still interested in how you don't get the SyntaxError.

PR #166, opened 2016-12-16
exposed sql native type via cursor->description

jstastny-cz:
I was missing a way how to get information about SQL datatype returned by the cursor. This way, the exposed SQL type code enables user to get type's info via cursor's getTypeInfo() method.

### DIFF
--- a/src/cursor.cpp
+++ b/src/cursor.cpp
@@ -321,14 +321,16 @@ static bool create_name_map(Cursor* cur, SQLSMALLINT field_count, bool lower)
             }
         }
 
-        colinfo = Py_BuildValue("(sOOiiiO)",
+        colinfo = Py_BuildValue("(sOOiiiOi)",
                                 (char*)name,
                                 type,                // type_code
                                 Py_None,             // display size
                                 (int)nColSize,       // internal_size
                                 (int)nColSize,       // precision
                                 (int)cDecimalDigits, // scale
-                                nullable_obj);       // null_ok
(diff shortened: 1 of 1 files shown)


mkleehammer:
I'm hesitant to do this since the DB API specification specially [defines description](https://www.python.org/dev/peps/pep-0249/#description) as a 7-item sequence.  It *seems* harmless to add one, but code like the following would fail:

```
a,b,c,d,e,f,g = cursor.description
```

That seems pretty unlikely, but if the next version of the specification defines an 8th item we'd be stuck.

I've been wanting to update this to use names like namedtuple or something.  (Or a Row?). If we did that, then we could recommend accessing it only by name ("sqltype"?) because it may have to be moved in the future.

Alternatively, we could put it either in a new description, such as descriptionex, or a new attribute all together like `types`.

Thoughts?

jstastny-cz:
I didn't realize the possible conflict with DB API.

The most tempting solution seems to be the one having a new attribute for this. If I may borrow a term from Java, something like ResultSetMetaData.

The attribute could be a dictionary, one of its entries called 'column_types' or something like that, so that it would be usable for other metadata-describing purposes.

The scope of the provided metadata could follow this section:
https://technet.microsoft.com/en-us/library/ms131297(v=sql.105).aspx

Can you assess the scope of such changes?

jstastny-cz:
I'll close this pr, for mentioned reasons.

Issue #167, opened 2016-12-17
Expose sql_type information about columns in resultset


jstastny-cz:
I am looking for an extension of metadata returned through Cursor.
For every column in a result set I'd like to be able to get sql_type code of that column.
Together with getTypeInfo method of cursor, user can get further information about the datatype.

findepi:
Same here. https://github.com/mkleehammer/pyodbc/wiki/Cursor#description docs suggests that it returns just that:
> 2. type code

but instead it returns Python type that values will be mapped to (e.g. `str`, `int`, `datetime.time` etc.)

@mkleehammer is there a way to get original (ODBC-level) type information?

jstastny-cz:
@findepi 
Hello,
there is a way. You can get it from the odbc driver. Actually, I've implemented the extension in my fork of this project, if you're interested.

findepi:
@jstastny-cz , is it `Cursor#colinfos` (https://github.com/jstastny-cz/pyodbc/blob/c278e74e44341a4333dbd2b471758b651abc7558/src/cursor.h#L112)?
Looks like that would be it! Are you planning on making a PR with this enhancement?

jstastny-cz:
@findepi 
Not that one.

I exposed the information in an additional field 'coldescription':
https://github.com/jstastny-cz/pyodbc/blob/2f703c07a98c70ac369e77603abd98e681204421/src/cursor.h#L117

and the PR is already in place (it never got into the project, as it is not project's priority I am afraid):
https://github.com/mkleehammer/pyodbc/pull/168

sundar-mudupalli-work:
Hi,

It would be really helpful to get this issue fixed and rolled into production. We are reading data from Microsoft SQL Server and need to know if the string column is 'char' or 'varchar'. Both show up as Python strings, the char value is blank padded, while the varchar is not.

Can the priority of this be bumped up.

Thanks.

Sundar Mudupalli

PR #168, opened 2016-12-17
Expose sql_type for columns in resultset #167

jstastny-cz:


mkleehammer:
@jstastny-cz This looks clean, but I'm wary about adding something to the interface without knowing what someone would do with it.  I'm not sure how one would make use of this information.  What was the use case for this?  Thanks.

jstastny-cz:
@mkleehammer Hello, we used pyodbc for ODBC interface/driver tests. What we needed to do was to check the actual SQL data type returned by the driver.

matthias-github:
If I understand the PR correctly, then we have the same use case. I am using pyodbc to generate SQL Server create tables DDL based on the result set. If the actual SQL type is returned, we can create exact DDL. Right now we are creating approximate DDL.

celestialorb:
Yes please! We have a use case that requires us to know the actual SQL type behind the column. We have been unable to come up with a clean workaround.

bkline:
@mkleehammer If the `Waiting for info` label is still appropriate, wouldn't it be appropriate to describe what information is still needed, beyond what has been provided in the responses above?

Issue #253, opened 2017-07-10
running on big endian systems


dangowrt:
### Environment

To diagnose, we usually need to know the following, including version numbers.  On Windows, be
sure to specify 32-bit Python or 64-bit:

- MIPS32r2 (big endian)
- Python 3.6.1
- pyodbc 4.0.17
- OS Linux (OpenWrt/LEDE)
- DB PostgreSQL 9.6.3
- driver psqlodbcw (and psqlodbca as well)

### Issue

While everything works nicely on all little endian systems, things break on big-endian boxes:
```
root@LEDE:/# python3
Python 3.6.1 (default, Jul 10 2017, 04:39:25) 
[GCC 5.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyodbc
>>> cnxn = pyodbc.connect("DRIVER={PostgreSQL Unicode};SERVER=localhost;DATABASE=template1;UID=postgres;PWD=password;PORT=5432")
>>> c = cnxn.cursor()
>>> c.execute("SELECT 'test' AS test")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
pyodbc.ProgrammingError: ('42601', '[42601] ERROR: syntax error at or near "匀䔀䰀䔀䌀吀\u2000✀琀攀猀琀✀\u2000䄀匀\u2000琀攀猀琀";\nError while executing the query (1) (SQLExecDirectW)')
>>> 
```
Note that ODBC generally works fine on the box, connecting to the same database using `isql` performing the same query is no problem:

```
root@LEDE:/etc# isql psqlodbc_test_dsn
+---------------------------------------+
| Connected!                            |
|                                       |
| sql-statement                         |
| help [tablename]                      |
| quit                                  |
|                                       |
+---------------------------------------+
SQL> SELECT 'test' AS test;
+-----+
| test|
+-----+
| test|
+-----+
SQLRowCount returns 1
1 rows fetched
```

Also note that on an otherwise identical little endian MIPS32r2 it works just fine, ie.
```
root@LEDE:/# python3
imporPython 3.6.1 (default, Jul 10 2017, 09:34:47) 
[GCC 5.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
        
>>> import pyodbc
>>> cnxn = pyodbc.connect("DRIVER={PostgreSQL Unicode};SERVER=localhost;DATABASE=template1;UID=postgres;PWD=password;PORT=5432")
[...]

PR #255, opened 2017-07-17
Use native endian for UTF-16

dangowrt:
Use UTF-16 encoding corresponding to the machine's native endian
as a default rather than setting 'utf-16le' even on big-endian
machines.
This change thus only affects big-endian platforms and results in
identical object code being generated on little-endian platforms.
As the default is also mentioned in the documentation, it should be
changed there as well once this patch has been merged.

Fixes https://github.com/mkleehammer/pyodbc/issues/253

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -18,6 +18,14 @@
 #include "cnxninfo.h"
 #include "sqlwchar.h"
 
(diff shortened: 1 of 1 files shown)


mkleehammer:
Somehow I missed this.  I apologize.

Issue #120, opened 2016-06-02
Slow Handling of executemany() 


dagostinelli:
There was a long standing, somewhat significant issue open at Google Code having to do with executemany() and MS SQL Server.   It does not yet appear to be fixed.  I'd like to call attention to it again. _(insert unhelpful chide for leaving open issues at Google Code and not bringing them over)_

[Issue #250](https://code.google.com/archive/p/pyodbc/issues/250)

Basically -- executemany() is taking forever.  The problem manifests itself when there are a lot of records (10k's or 100k's) and using MS-SQL Server.

When one uses the SQL Profiler, you can see, for every statement, an `exec sp_prepexec`, followed by an `exec sp_unprepare`.  

Here's a bit of sample code that makes it happen.

```
import pyodbc
dest_conn = pyodbc.connect('DSN=MyDSNToASQLServer;UID=someuser;PWD=password;')
dest_cursor = dest_conn.cursor()

sql = 'INSERT INTO MyTable (Name, Address, Phone) VALUES (?, ?, ?)'

data = []

## OMITTED
# Fill data[] with a lot of data 10k's of rows or more
##

# this takes forever when there are a lot of records
dest_cursor.executemany(sql, data)
dest_conn.commit()

```

keitherskine:
See issue #62 .  I believe one of the problems is that "parameter arrays" require the input data to be converted to a fixed-width encoding, i.e. NOT utf-8 or utf-16, because the data has to be temporarily stored in a block of memory, in grid-like fashion.  This means converting the input data to something like UCS-16 (with a small but significant possibility of data loss/corruption), and having to figure out how big the grid should be.  Not an easy problem to solve, but it would certainly help me in the work I do.

dagostinelli:
Is there a workaround that we can use for now?  

Someone mentioned using a transaction in the Google Code link -- I'm using that and it doesn't help.  In Issue #62, it was mentioned to load 250 records at a time.  I can try that.  

Did anything work for you?

keitherskine:
I use the bcp bulk load utility - a lot.  Even for routine updates and inserts, I write the updates to a file, bcp them to a temporary table, then update the target table in one transaction.  It's a very roundabout way of doing updates, but if you're doing millions of updates to tables containing hundreds of millions of rows, it's the fastest way.  By the way, I wrote python modules to do all the heavy lifting (which are company-proprietary and I can't publish - sorry).

[...]

PR #256, opened 2017-07-19
Parameter arrays for faster executemany()

v-chojas:
Fix for https://github.com/mkleehammer/pyodbc/issues/120 using parameter arrays.
This feature defaults to off (using the existing execute()-in-a-loop) but may be enabled by setting Cursor.fast_executemany = True.

### DIFF
--- a/src/cursor.cpp
+++ b/src/cursor.cpp
@@ -904,24 +904,27 @@ static PyObject* Cursor_executemany(PyObject* self, PyObject* args)
             PyErr_SetString(ProgrammingError, "The second parameter to executemany must not be empty.");
             return 0;
         }
-#if 0
-        for (Py_ssize_t i = 0; i < c; i++)
+        if (cursor->fastexecmany)
         {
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -50,19 +53,21 @@ def finalize_options(self):
...
 
+
 def main():
+
     version_str, version = get_version()
 
     settings = get_compiler_settings(version_str)
 
-    files = [relpath(join('src', f)) for f in os.listdir('src') if f.endswith('.cpp')]
+    files = [ relpath(join('src', f)) for f in os.listdir('src') if f.endswith('.cpp') ]
```
mkleehammer: Small suggestion here - I'm moving to using no spaces (like it was before) for PEP8 compliance.  I'm now using flake8 (via flycheck in Spacemacs) which flags these.

mkleehammer: Wait - this may be because it just needs to be rebased with master.  I *recently* cleaned up lint / flake8 warnings.

gordthompson:
Also relates to #62

Issue #213, opened 2017-03-14
Need a way to define data type/length when passing variables in SQL Server


ZaM-Rai:
I'm doing selects/inserts/updates on my tables and SQL Server is creating an execution plan for every permutation of character lengths that the selects/inserts/updates into the table from python.

I believe this is causing some issues with the performance of the db and may create some memory issues.

I would like to be able to set the data type/length when passing parameters. Ideally, pyodbc would know the datatype/length before inserting and set them accordingly but I don't know if this will create more issues issues than it will solve, so being able to define them myself might help.

More specifically, when I run a trace, this is what I see:

```
declare @p1 int
set @p1=4882452
exec sp_prepexec @p1 output,N'@P1 nvarchar(11),@P2 nvarchar(16)',N'SELECT TOP 1 C_KEY FROM C WHERE A_NM = @P1 AND C_NM = @P2',N'Zantesuken1',N'RasalhagueLegacy'
select @p1
```

and when looking at the execution plans, I find that an execution plan has been created for every single permutation of these selects.

Perhaps something like this:

cursor.execute("SELECT X FROM Y WHERE Z1 = ? AND Z2 = ?", param1.type(varchar(50)), param2.type(bigint))

No idea if this makes sense and maybe this is already possible somehow and I don't know about it, so let me know if there's a way around this issue.

*edit* Also noticed that the auto-incrementing @p1 seems to be going up really quickly (I'm doing a lot of transactions). Is there any way to reset this counter? I've tried deleting the cursor, closing the connection, but nothing seems to be working.

mkleehammer:
Interesting.  I guess it makes sense if the target size will matter to the execution plan.  I need to look into this for one of my own projects.

Is it possible to test the impact using casts?  Something like `Z1 = cast(? as varchar(50))`.

As for @p1, I actually don't know why the driver does that, but I don't believe it will be a problem.  I have some 24x7 applications using pyodbc pretty heavily and they usually open a connection and leave it open for weeks or months - they reset when there is an error or it is restarted.  I haven't noticed anything related to it.

gordthompson:
@ZaM-Rai re: closing connection - The ODBC infrastructure on Windows automatically enables connection pooling for the SQL Server ODBC drivers, so you could try `pyodbc.pooling = False` before opening your connection(s), or even disable connection pooling for your SQL Server ODBC driver via the ODBC Administrator (odbcad32.exe).

gordthompson:
[...]

PR #257, opened 2017-07-19
A preliminary implementation of Cursor.setinputsizes()

v-chojas:
To help fix https://github.com/mkleehammer/pyodbc/issues/213 .

### DIFF
--- a/src/cursor.cpp
+++ b/src/cursor.cpp
@@ -437,7 +437,7 @@ static void Cursor_dealloc(Cursor* cursor)
     {
         closeimpl(cursor);
     }
-
+    Py_XDECREF(cursor->inputsizes);
     PyObject_Del(cursor);
 }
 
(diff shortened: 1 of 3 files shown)


mkleehammer:
I screwed up and merged one of your later pull requests so I had to do some merging manually.  This has been merged, but won't show it here.

mkleehammer:
FYI: I'm going to open a new pull request for this.  I merged it after some very simple tests but should have spent a bit more time.  There is a lot of duplication between the single statement and multi-statement versions which makes it difficult to maintain and leads to different behavior.  One current example is that encodings are ignored if `sizeof(PY_UNICODE) == sizeof(wchar_t)` (or something like that - I'm not looking at the code).

I see I can reopen this pull request, but I don't know what that would mean since it has already been merged, so I'll create a new one.

Issue #251, opened 2017-06-22
Decoding of error messages returned by server


gordthompson:
(Prompted by [this Stack Overflow question](https://stackoverflow.com/q/44690036/2144390).)

This code

```python
import pyodbc
import sys
print("Python " + sys.version)
print("pyodbc " + pyodbc.version)
conn_str = (
    "Driver=ODBC Driver 11 for SQL Server;"
    "Server=192.168.1.142,49242;"
    "Database=myDb;"
    "Language=Japanese;"
    "Trusted_Connection=yes;"
)
cnxn = pyodbc.connect(conn_str)
crsr = cnxn.cursor()
try:
    crsr.execute("SELECT * FROM nada")  # table does not exist
except pyodbc.ProgrammingError as pe:
    print(pe)
```

produces this when run in PyCharm under Windows

```
Python 3.6.0 (v3.6.0:41df79263a11, Dec 23 2016, 07:18:10) [MSC v.1900 32 bit (Intel)]
pyodbc 4.0.17
('42S02', "[42S02] [Microsoft][ODBC Driver 11 for SQL Server][SQL Server]??????? 'nada' ?????? (208) (SQLExecDirectW)")
```

Wireshark shows that the server is returning the error message as UTF-16LE

![packet](https://user-images.githubusercontent.com/15894006/27446860-2d57093c-573c-11e7-86ad-1e3c25501cc5.png)

and

```python
print("\u30AA\u30D6\u30B8\u30A7\u30AF\u30C8\u540D 'nada' \u304C\u7121\u52B9\u3067\u3059\u3002")
```

displays this

> オブジェクト名 'nada' が無効です。

which Google Translate tells me means

> Invalid object name 'nada'.

I tried tweaking the encoding settings as described in [the Unicode wiki page](https://github.com/mkleehammer/pyodbc/wiki/Unicode), but they did not seem to make a difference.

Is this something that pyodbc can work around, or is it simply a limitation of the ODBC infrastructure ("driver manager"?) on Windows?

michaelfindlater:
Hi there,

I'm the original poster of the Stack Overflow comment. 

We're experiencing the this exact issue. Even if Japanese is set, the server response (UTF-16LE) comes back garbled. `Language=English` obviously works fine and results come back in English.

Cheers,

Michael

gordthompson:
Additional information: When I run the same test code under Xubuntu 16.04 I get

~~~
[...]

PR #258, opened 2017-07-24
Retrieve Unicode diagnostic messages

v-chojas:
To fix https://github.com/mkleehammer/pyodbc/issues/251 diagnostic messages are retrieved using the Unicode API SQLGetDiagRec**W**. The default is UTF-16LE, as usual for ODBC, but can be configurable when a connection is provided since it is based on connection.unicode_enc.

Due to the amount of duplication I noticed while preparing this fix, in the future it would be good to revisit pyodbc's error handling in general, since a lot of it could be factored out.

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -81,7 +81,7 @@ static bool Connect(PyObject* pConnectString, HDBC hdbc, bool fAnsi, long timeou
         ret = SQLSetConnectAttr(hdbc, SQL_ATTR_LOGIN_TIMEOUT, (SQLPOINTER)(uintptr_t)timeout, SQL_IS_UINTEGER);
         Py_END_ALLOW_THREADS
         if (!SQL_SUCCEEDED(ret))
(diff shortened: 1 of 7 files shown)


mkleehammer:
This is fantastic.  Thanks for this.  I'm really looking forward to working with Luis and yourself.

michaelfindlater:
Thank you to everyone who worked to fix this.

mkleehammer:
I need to update this as it breaks all non UTF16 drivers.  (This is the same issue we have with the code binding parameters as arrays.)  Do remember that a lot of drivers *do not* follow the UTF16 / SQLWCHAR specification.  I've been annoyed with them over the years, but it isn't going away.  That's why I added all of the encoding options in 4.0.

When something is added, we need to make sure there is a unit test for SQL Server, which does follow the spec nicely, and PostgreSQL which is almost always installed to use UTF8.

mkleehammer:
Updated in 82770734e845fdd1137762e250461708f837d736.  I should have put a reference to this ticket in the message.  (I've been working on so many different systems lately that I've been making small mistakes like that.)

PR #259, opened 2017-07-31
Support non-integer-valued connection attributes

v-chojas:
Enables use of string and pointer connection attributes. Type mappings are as follows:

Python object | Value type
----------|---------------
buffer(2.x) | SQL_IS_POINTER
bytearray(>=2.6)|SQL_IS_POINTER
bytes(2.x)|string length
bytes(3.x) |SQL_IS_POINTER
integers| SQL_IS_INTEGER or SQL_IS_UINTEGER depending on sign
unicode| string length

Another extension I added is to allow sequence of values to be used, which causes the SQLSetConnectAttr call to be repeated. Some drivers may find this useful for their configuration interfaces. E.g. [SQL_COPT_SS_CEKEYSTOREDATA](https://docs.microsoft.com/en-us/sql/connect/odbc/using-always-encrypted-with-the-odbc-driver#working-with-column-master-key-stores) can be set with `attrs_before = { 1252: [ data1, data2, data3 ] }`  (that particular one is more useful post-connect, but the same principle applies.)


v-chojas:
unixODBC, or at least the version your Travis is using, doesn't have that defined because it's only in ODBC 3.8+. Further investigation reveals that, according to 
https://docs.microsoft.com/en-us/sql/odbc/reference/develop-driver/upgrading-a-3-5-driver-to-a-3-8-driver , "SQL_ATTR_RESET_CONNECTION is only used in communication between the Driver Manager and a driver. An application cannot set this attribute directly." so it is not useful to pyodbc, and I will remove it.

mkleehammer:
Thanks.  BTW, I merged in your other pull requests so this one will need some minor cleanup due to the RaiseError API change.  (I also majorly screwed up by merging that one first.  That's what I get for hurrying.)

v-chojas:
I merged your other merges back into this one, so it should be ready to merge now.

v-chojas:
Any updates on merge-ability of this?

v-chojas:
I fixed some merge conflicts, how is it now?

v-chojas:
Any update on merging this? It is needed to support some features of ODBC Driver for SQL Server, such as Azure Active Directory Token authentication and Always Encrypted.

v-chojas:
Leave a comment Can this be merged now?

mkleehammer:
The commit portion shows "This branch cannot be rebased due to conflicts".  I assumed everyone could see that.  Sorry.  Can you rebase to the current master?

v-chojas:
How about now?

v-chojas:
It looks OK from here:
```
This branch has no conflicts with the base branch
Only those with write accessLearn more about permission levels to this repository can merge pull requests. 
``
Are you able to merge it?

v-makouz:
Any news on this?

v-chojas:
Can this be merged?

mkleehammer:
FYI: This doesn't compile with clang on macOS...

    src/connection.cpp:134:25: error: ordered comparison between pointer and zero ('SQLPOINTER' (aka     'void *') and 'int')
            vallen = ivalue < 0 ? SQL_IS_INTEGER : SQL_IS_UINTEGER;

I'll fix it, but we need to make sure things get compiled with multiple compilers.  (VS usually has the best error checking and messages, but clang is pretty good lately.)

mkleehammer:
@v-chojas Isn't this just asking for more Unicode problems?

    if (sizeof(Py_UNICODE) == 2) // Should be compile-time.
    {
            ivalue = PyUnicode_AS_UNICODE(value);
    }
    else
    {
            Object uniStr(PyUnicode_EncodeUTF16(
                              PyUnicode_AS_UNICODE(value),
                              PyUnicode_GET_DATA_SIZE(value),
                              0,
                              -1));
            ivalue = PyBytes_AS_STRING(uniStr.Get());
    }

We have a bunch of encodings configured, such as metadata, etc., and we're not using them.  We assume that Python Unicode is always 2-byte - either `sizeof(Py_UNICODE) == 2` or we use UTF-16.

v-chojas:
Leave a commentOK, I will take a look at using the specified encoding.

v-chojas:
@mkleehammer  I finally had some time to revisit this and think I made the changes so that it is now using the optional `encoding=` parameter of the `connect()`. I couldn't use the configured encodings in the connection object, because at that point it doesn't even exist yet.

> I'll fix it, but we need to make sure things get compiled with multiple compilers. (VS usually has the best error checking and messages, but clang is pretty good lately.)

It did compile for me on clang/macOS 10.11. I've also tested it on Linux and Windows (VS2013).

PR #271, opened 2017-08-16
Add error subclass mappings for more SQLSTATEs

bobince:
Detect more operational error codes so we can raise more specific error classes for these known cases (seen coming from MSSQL — I can't vouch that these offer any improvement for other DBMSs).

### DIFF
--- a/src/errors.cpp
+++ b/src/errors.cpp
@@ -19,15 +19,27 @@ struct SqlStateMapping
 
 static const struct SqlStateMapping sql_state_mapping[] =
 {
+    { "01002", 5, &OperationalError },
+    { "08001", 5, &OperationalError },
+    { "08003", 5, &OperationalError },
+    { "08004", 5, &OperationalError },
+    { "08007", 5, &OperationalError },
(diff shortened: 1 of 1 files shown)


src/errors.cpp
```diff
@@ -19,15 +19,27 @@ struct SqlStateMapping
 
 static const struct SqlStateMapping sql_state_mapping[] =
 {
+    { "01002", 5, &OperationalError },
+    { "08001", 5, &OperationalError },
+    { "08003", 5, &OperationalError },
```
mkleehammer: 08003 is "Connection not open".  Wouldn't that be ProgrammingError?

bobince: When we've had it, it has been the result of executing a query on a connection that has unexpectedly closed at the server end (eg due to a timeout), which feels more like an operational problem.

(Although, it's unclear whether this is what the sqlstate is *supposed* to represent, as the definitions are pretty fuzzy and don't necessarily match reality...)

mkleehammer: Well then, let's leave them as-is.  Thanks.

PR #278, opened 2017-09-05
Add /usr/local/lib to library_dirs linker setting for OS X

neilodonuts:
This line fixes the ability to install Pyodbc through PIP after building unixODBC from source on OS X El Capitan and beyond. Otherwise PIP fails with "ld: library not found for -lodbc".


mkleehammer:
Great.  Thanks!

Issue #260, opened 2017-08-01
VarChar and LVarChar Handling with IBM Informix


a-musing-moose:
### Environment

To diagnose, we usually need to know the following, including version numbers.  On Windows, be
sure to specify 32-bit Python or 64-bit:

- Python: 3.6.2
- pyodbc: 4.0.17
- OS: Linux & MacOS
- DB: Informix
- driver: Informix ODBC driver from IBM

### Issue
When attempting to insert more than 255 characters into a Varchar field we get the following error: 

`('HY000', '[HY000] [Informix][Informix ODBC Driver][Informix]No cast from text to lvarchar. (-9634) (SQLParamData)')`

It seems that a quirk of Informix is that all varchar fields require data to be sent as the `SQL_VARCHAR` type (see: https://www.ibm.com/support/knowledgecenter/en/SSGU8G_12.1.0/com.ibm.odbc.doc/ids_odbc_083.htm) irrelevant of length.  Only Text fields need to be sent as `SQL_LONGVARCHAR`.

I'm not a C programmer, but we managed to adapt the code to make it work for us by really just using `SQL_VARCHAR`. But clearly this breaks both `Text` fields and likely everyone else use of PyODBC which is far from ideal.

mkleehammer:
I think a solution to #213 would help you here.  I've been holding off on that because I don't have a good "Pythonic" way to pass it, but I think it is time to come up with something.

Do you have *something* working for you for now?

highway900:
No unfortunately this doesn't do the trick.
e.g.
`cursor.execute('insert into t_temp(lvarname) values(cast(? as lvarchar(5000)))', 'p'*256)` still returns the `No cast from text to lvarchar. ` error
`cursor.execute('insert into t_temp(lvarname) values(cast(? as varchar(5000)))', 'p'*256)` returns `Maximum varchar size has been exceeded` which should be right. But our situation is interesting due to the quirk mentioned in the issue.
`pyODBC` will use `GetColumnSize` in `cnxninfo.cpp` to query the max length of SQL_VARCHAR from the informix ODBC driver, this will always return 255.

v-chojas:
Yes, this is definitely another case where parameters need to be bound with the exact SQL type the server is expecting, despite the fact that to Python there is no difference and they are the same (i.e. string) types. Extending setinputsizes() to specify the type too seems like the way to solve this without resorting to potentially slow SQLDescribeParam (not sure what its performance is like with Informix.)

[...]

PR #280, opened 2017-09-07
Extend setinputsizes() to allow specifying type+length+precision/scale

v-chojas:
This change extends setinputsizes() in the following way: each element of the sequence passed to it can be, in addition to a single integer, a sequence of `[type, length, precision/scale]` to cause the specified parameter to always be bound and sent using that type. The type can be one of the `pyodbc.SQL_*` constants, while the length and precision/scale are integers. Any of them can be omitted or `None`'d to not override the corresponding value. Specifying a single integer will, as before, only override the length. E.g., the following statement:

`cursor.setinputsizes([ 15, [ pyodbc.SQL_INTEGER ], None, [ None, 12, 5 ], [ pyodbc.SQL_VARCHAR, 3000 ] ])`

Has the following effects:
- Column 1 will always be bound with a length of 15.
- Column 2 will always be sent with type SQL_INTEGER.
- Column 3 will not have any aspects of its type overridden.
- Column 4 will always be sent with length 12 and precision/scale 5, while its SQL type will be autodetected from the Python type.
- Column 5 will always be sent as a SQL_VARCHAR of length 3000.
- Columns 6 and above will not have any aspects of their type overridden.

Relevant issues: https://github.com/mkleehammer/pyodbc/issues/260 https://github.com/mkleehammer/pyodbc/issues/213

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-import sys, os, re
+import sys, os, re, platform
 from os.path import exists, abspath, dirname, join, isdir, relpath
 
 try:
@@ -12,13 +12,16 @@ except ImportError:
 from distutils.extension import Extension
 from distutils.errors import *
 
-OFFICIAL_BUILD = 9999
+if sys.hexversion >= 0x03000000:
+    from configparser import ConfigParser
+else:
+    from ConfigParser import ConfigParser
 
+OFFICIAL_BUILD = 9999
 
 def _print(s):
     # Python 2/3 compatibility
     sys.stdout.write(s + '\n')
-
 
 class VersionCommand(Command):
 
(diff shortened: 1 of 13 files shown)


mkleehammer:
This was an ambitious change.  Thanks!

I often add to the DB API specification, but I'm a bit concerned about deviating from it too much.  We should consider that.

Also, I really hate to add *anything* else to the parameters until they are fixed up.  I completely abrogated responsibility a few months back and merged in the fast executemany with a cursory test but without going over the code.  There are now *two* sections of code for binding parameters for each type and they do not use the same rules at all.  For example, the string binding assumes every driver is using UCS2 which is absolutely not the case.  Even though it is what the specification says, I spent a lot of effort writing version 4.x to add encoding support so it works with all databases.

Similarly, Decimal support was set to a string because so few databases get the binary part right the last time I tried.  (Including well known ones like Oracle, IIRC.)

I started a branch to consolidate, but I was derailed by a bunch of other projects and would need to start fresh.  I'm not sure the best place to have a conversation about that.

Do you see a relatively easy way to first consolidate the memory handling for binding parameters between the array and non-array versions?

v-chojas:
Closing this since the functionality is included in the parameter array refactoring.

PR #279, opened 2017-09-06
Make ColumnSize be size in characters after encoding

v-chojas:
Fixes https://github.com/mkleehammer/pyodbc/issues/221

Issue #221, opened 2017-03-21
MSSQL: string with large unicode character is truncated on insert


dirkjonker:
When inserting a string containing emoji characters into a `NVARCHAR` column in MSSQL using FreeTDS 1.00.26 + unixODBC 2.3.4 + Python 3.5 + pyodbc 4.0.15 and the default connection properties (as per recommendation in the documentation):

```
>>> text = 'Test 1 2 3 😂  foo bar'

>>> cursor.execute("CREATE TABLE dbo.unicode_test (val NVARCHAR(100));")

>>> cursor.execute("INSERT INTO dbo.unicode_test (val) VALUES (?);", (text,))

>>> cursor.execute("SELECT TOP (1) val FROM dbo.unicode_test;")

>>> print(cursor.fetchone())
('Test 1 2 3 😂  foo ba', )
```

As you can see, the final "r" is dropped. For each extra emoji in the text, more characters seem to be truncated (N emoji = last N characters are dropped).

This seems to happen when pyodbc is converting the UTF8 string when executing the INSERT statement, since the database does not contain the final character (I checked this with other tools). 

When I convert the text to UTF-16-LE using `text.encode('utf_16_le')`, there is no truncation.

gordthompson:
I am unable to reproduce your issue under PyCharm Community Edition 2016.3.3 with Python 3.6 and pyodbc 4.0.15. My code ...

```python
import pyodbc
print(pyodbc.version)
cnxn = pyodbc.connect(
    r"DRIVER=ODBC Driver 11 for SQL Server;"
    r"SERVER=(local)\SQLEXPRESS;"
    r"DATABASE=myDb;"
    r"Trusted_Connection=yes;")
crsr = cnxn.cursor()

text = 'test 1 2 3 😂 foo bar'

crsr.execute("CREATE TABLE #unicode_test (val NVARCHAR(100))")
crsr.execute("INSERT INTO #unicode_test (val) VALUES (?)", text)

value = crsr.execute("SELECT TOP 1 val FROM #unicode_test").fetchval()
print(value)

crsr.close()
cnxn.close()
```

... produces ...

```none
C:\Users\gord\PycharmVirtualenvs\py3pyodbc\Scripts\python.exe C:/Users/gord/PycharmProjects/pyodbc_issue_221/main.py
4.0.15
test 1 2 3 😂 foo bar

Process finished with exit code 0
```

I also get the same results if I use a Unicode escape instead of the emoji character

```python
[...]

PR #287, opened 2017-09-21
Make ColumnSize be size in characters after encoding

v-chojas:
Trying https://github.com/mkleehammer/pyodbc/pull/279 again.

### DIFF
--- a/src/params.cpp
+++ b/src/params.cpp
@@ -729,10 +729,6 @@ static bool GetUnicodeInfo(Cursor* cur, Py_ssize_t index, PyObject* param, Param
 
     info.ValueType = enc.ctype;
 
-    Py_ssize_t cch = PyUnicode_GET_SIZE(param);
-
-    info.ColumnSize = (SQLUINTEGER)max(cch, 1);
-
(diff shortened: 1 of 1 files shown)


mkleehammer:
Do we still want to include this or wait for the bigger changes we discussed?

Also note that we should use ODBCCHAR instead of SQLWCHAR with sizeof due to some incorrect headers either in unixODBC or iODBC.  One of them uses a 4-byte data type but it is actually 2-byte internally.  I should try to find the details for that.

iamsed:
Merging it would fix the "ODBC Driver 13 for SQL Server" under Linux [ref](https://github.com/mkleehammer/pyodbc/issues/221#issuecomment-288247441)

So maybe it could be included "in the meantime" if you guys decide to work on the bigger changes.

mkleehammer:
I assume we can delete this now?

v-chojas:
Yes, this was a fix to the old code which has changed significantly since the parameter array refactoring. ColumnSize is now retrieved from the server via SQLDescribeParam.
Leave a comment

PR #298, opened 2017-11-07
use odbc_config if it's present

pkit:
some/most odbc implementations on *nix-like systems use `odbc_config`
as a source of compile-time options about the odbc installation
this change will try to use `odbc_config` at compile time to correctly set LDFLAGS and CFLAGS


mkleehammer:
Fantastic.  I'm tempted to extract that and call it on macOS also.  Thanks.

PR #301, opened 2017-11-22
Parameter Array Refactoring

v-chojas:
These changes combine both the execute()/regular executemany() with the fast executemany() parameter handling code as much as possible, reducing the duplication from before; there has been some minor enhancement too:

- SQLDescribeParam() used to detect server's parameter type
- setinputsizes() can override the type detection and force a specific parameter type
- Data type conversion has been enhanced to expand the range of usable values; for example, a long integer can be inserted into a character column with full precision. This is done by using the SQL type to decide which C type the Python object should be converted to when binding.
- SQL Server extended date/time types can be inserted at full precision
- datetime objects inserted into SQL Server datetimeoffset columns will preserve their timezone information
- `cursor.decimal_as_string` added to force string binding 
- fast executemany() will perform character encoding conversion if requested by encoding settings.
- fixed small buffer overflow bug in cursor.noscan (wrong type for attribute causes out-of-bounds write on 64-bit systems)

### DIFF
--- a/src/cursor.cpp
+++ b/src/cursor.cpp
@@ -27,6 +27,9 @@
 #include "sqlwchar.h"
 #include <datetime.h>
 
+bool ParamSetup(Cursor *cur, PyObject *sql, PyObject* original_params, bool skip_first);
+int BindAndConvert(Cursor *cur, Py_ssize_t i, PyObject *cell, ParamInfo *ppi);
+
 enum
 {
     CURSOR_REQUIRE_CNXN    = 0x00000001,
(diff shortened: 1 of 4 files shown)


mkleehammer:
I'm not ignoring this, I just wanted to take some time to review and test properly.

If you compile on macOS, the pgtests for Python 3.6 have 2 failures and 12 errors I need to look at.  I haven't tested on Linux or with MySQL yet, but we need to do that.  This needs to not break those builds.

Also, in macOS I'm getting two warnings we should eliminate.  I'm guessing Visual C++ is emitting them too, but haven't built it there yet: `if (ret = PyLong_Check(obj))`.  I'm trying to eliminate all compiler warnings so real findings are not ignored.  We can add another set of parens to eliminate this with GCC.  I think VC recognizes the same.

src/cursor.cpp
```diff
@@ -661,88 +777,8 @@ static PyObject* execute(Cursor* cur, PyObject* pSql, PyObject* params, bool ski
...
-                    if (!SQL_SUCCEEDED(ret))
-                        return RaiseErrorFromHandle(cur->cnxn, "SQLPutData", cur->cnxn->hdbc, cur->hstmt);
-                }
-            }
-#endif
-            ret = SQL_NEED_DATA;
-        }
-    }
+    if (!ProcessDAEParams(ret, cur, false))
+        return 0;
```
mkleehammer: Since the previous error handler on line 776 freed the parameter data, I bet we should do so too.  Unless it does it in PDP, but then the next line would be confusing.

src/cursor.cpp
```diff
@@ -27,6 +27,9 @@
 #include "sqlwchar.h"
 #include <datetime.h>
 
+bool ParamSetup(Cursor *cur, PyObject *sql, PyObject* original_params, bool skip_first);
+int BindAndConvert(Cursor *cur, Py_ssize_t i, PyObject *cell, ParamInfo *ppi);
```
mkleehammer: Two things - first I think it would be better if these were in the header file.  I get we only use it here.

More importantly, let's change `int` to `bool` since it returns true and false.

mkleehammer:
OK, I found out one of the issues I'm having.  Something in PyToCType is failing so it returns false, however, it doesn't always set a Python exception.  That's causing the following:

    SystemError: <built-in method execute of pyodbc.Cursor object at 0x10402a918> 
      returned NULL without setting an error

For each function we should decide, and document usually, if it either (1) always sets an exception or (2) never sets an exception.

I'll keep digging into why it is failing exactly, but I wanted to record this so we don't forget *before* I fix the lower problem.

mkleehammer:
It might be good if we pull out some chunks of PyToCType into inlined functions.  With C++ and modern C compilers we get the benefit of smaller, easier to digest functions *and* the performance of inlined code.

mkleehammer:
OK, the first issue is tests3/pgtests.py test_fetchval fails.  It is writing a single Python 3 `str` using the "utf-8" encoding:

    self.cnxn.setencoding(encoding='utf-8')
    ...
    expected = "test"
    self.cursor.execute("insert into t1 values(?)", expected)

This fails with the error above about NULL with no error because the detect code looks at the encoding and decides it is SQL_C_CHAR:

    pi->ValueType = cur->cnxn->unicode_enc.ctype;

 but the PyToCType Unicode handling starts with:

    if (pi->ValueType != SQL_C_WCHAR)
        return false;

v-chojas:
Are you ready for me to start making review changes? Let me know when you're done with this first round of review.

mkleehammer:
I'd prefer it if you could test these changes against at least PostgreSQL and MySQL.  The existing tests already checked in are a good start, and we can add fastexecutemany=True pretty easily.

Do you have access to a Linux VM?  Setup of both of those databases is a breeze and running the tests is very fast.

v-chojas:
We are in progress of testing those databases and making changes as necessary to fix any issues discovered.

mkleehammer:
Fantastic.  That was a much smaller change than I anticipated.  Merged in.  I'll make a release as soon as I see if there are any other urgent fixes or merge requests that can go in.

mkleehammer:
@v-chojas Argh - I might have proceeded too quickly.  On Windows with 32 bit Python 2.7 and Python 3.6, the text_xxx tests fail with a truncation issue.  I haven't dug in yet but from the error it looks like the buffer size is being set to a negative that is cast to an unsigned value.

Are you seeing these errors with the sqlserver unit tests?  (I'm assuming you run those.)

I've merged to master, so you can run there.  It could be something introduced after, but most of those are related.

v-makouz:
I've just tried it and I see the same issue, 64 bit python tests are fine but on 32 bit (I tested with 2.7) I see the truncation errors, I will investigate further and see what I can find

v-makouz:
Upon investigating this, I think this is what the root of the problem is: the new parameter array code uses SQDescribeParam to get BufferLength (among other things), and for the SQL "text" the BufferLength is always 0x7FFFFFFF and since the code is dealing with SQLWCHARs it multiplies it by two to get 
0xFFFFFFFE, which is -2 in 32 bit. The tests fail on a check that is easy enough to fix by casting the BufferLength to unsigned long, but the real problem is SQLBindParam which will fail due to the buffer being set to -2.

I'm not sure what is the best solution to this, either we need to handle "text" types specially and set the buffer to zero when binding them or some generic handling of BufferLength before calling SQLBindParam to make sure it's not negative.

Both v-chojas and I will be out till next year, so I'm afraid that neither of us is likely to be able to solve this before January

PR #304, opened 2017-11-30
Fix memory leak in fast executemany()

v-chojas:
Fix a small but significant memory leak with fast executemany().

You can merge this one to fix the leak, if you are not ready to accept the parameter array refactoring.


vertigo235:
Is it possible to apply this fix to 21?

PR #324, opened 2018-01-05
Support for SQL INTERVAL Types by mapping to two distinct Python types

padhia:
SQL `INTERVAL` types are classified into two distinct groups -- one that has **month** as the most granular time period and the other which has **microsecond** as the most granular time period. Per the SQL standard, it is meaningless to convert from one group of interval types to the other and thus is not permitted.

This PR provides support for these two groups of SQL `INTERVAL` types by mapping them to Python `int` (number of months) and `datetime.timedelta` respectively as their canonical representations.

### DIFF
--- a/src/getdata.cpp
+++ b/src/getdata.cpp
@@ -644,6 +644,55 @@ static PyObject* GetDataTimestamp(Cursor* cur, Py_ssize_t iCol)
 }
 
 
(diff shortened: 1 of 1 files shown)


mkleehammer:
Nice job coming up with a reasonable interface: month & microsecond.

Can you add some unit tests for at least SQL Server and PostgreSQL for these before we merge?

PR #327, opened 2018-01-10
Fix invalid cursor state upon executemany()

v-chojas:
Free the previous results before attempting to execute again on the cursor.


gaborbernat:
@mkleehammer any update on this, this is kinda critical

PR #352, opened 2018-02-20
Revert parameter code

mkleehammer:
Back out the parameter refactoring and do it again.  We should make tests for any problems that were reported so we know we don't reintroduce them.


v-chojas:
Your AppVevor appears to be stuck.

PR #357, opened 2018-02-27
Fix use of Unicode connection string in Python2

v-chojas:
If a Unicode connection string is used, hash.update throws an exception about invalid encoding.

### DIFF
--- a/src/cnxninfo.cpp
+++ b/src/cnxninfo.cpp
@@ -46,6 +46,14 @@ static PyObject* GetHash(PyObject* p)
     if (!bytes)
         return 0;
     p = bytes.Get();
+#else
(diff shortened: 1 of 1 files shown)


mkleehammer:
This forces everything to UTF8.  Do any databases expect latin1, where extended characters would completely change if we encoded as UTF-8?  I don't want to rush something as important as the connect method.

One problem with setting the metadata encodings on the connection is you have to have a connection first.  I'd like to use a "context" like the SSLContext, etc., since that seems to be becoming a Python standard, and provide a connect method there.  Perhaps pyodbc 5.x

v-chojas:
As far as I can see, this is only used internally for connection string hashing and doesn't ever reach any ODBC calls itself.

mkleehammer:
Ha!  I completely missed that.  Thanks.

PR #374, opened 2018-04-16
Use correct size for SQL_ATTR_NOSCAN attribute

v-chojas:
Prevents the buffer from overflowing (and crashing Python).

From https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqlsetstmtattr-function

"SQL_ATTR_NOSCAN (ODBC 1.0) An **SQLULEN** value that indicates whether the driver should scan SQL strings for escape sequences"


v-chojas:
Can this be merged since it fixes a crash on some Python versions?

Issue #380, opened 2018-04-27
String right truncation pyodbc.ProgrammingError when using `fast_executemany`


seanfdnn:
### Environment

To diagnose, we usually need to know the following, including version numbers.  On Windows, be
sure to specify 32-bit Python or 64-bit:

- Python: 3.6.3
- pyodbc: 4.0.23
- OS: Debian Stretch
- DB: SQL Server 
- driver: Microsoft ODBC Driver 17 for SQL Server libmsodbcsql-17.0.so.1.1

### Issue

When executing a parameterized "UPDATE" statement with `fast_executemany`, the following exception is thrown.

```
ERROR - encoding with 'utf-16le' codec failed (SystemError: <built-in function utf_16_le_encode> returned a result with an error set)              
pyodbc.ProgrammingError: ('String data, right truncation: length 666 buffer 510', 'HY000')
```

The column widths in the target database have more than enough size to hold the data. 
This seems equivalent to Issue #337  except that in this case, I'm using 4.0.23 of pyodbc and MS ODBC Driver 17, which fixed the issue in that post but continue to be a problem for me.

I've trimmed down the data to just two rows and can consistently reproduce the problem with just those two rows. 

I'm using SQL Alchemy to construct the statements and am using the `connection.execute(stmt, [row1, row2])` method to execute.

Turning `fast_executemany` off fixes the problem, but unacceptably impacts performance.

seanfdnn:
Some more details:
The column causing the issue is VARCHAR(2000)

After reading #295 , I considered that, even though a TEMP table is not the target of the insert, there is one used in the statement. Changing to a global temp table with ## instead of # fixes in this specific instance, but then the same error happens again on a different table, different column with a different set of data. The TEMP table does not use any varchars, only one integer column, and everywhere in the DB, all columns are varchar, not nvarchar

```
SET NOCOUNT ON;
BEGIN TRY;
BEGIN TRANSACTION
-- Execute statement here
IF @@ROWCOUNT = '0' THROW 5100, 'FAiled to update or delete record'
COMMIT 
END TRY
BEGIN CATCH
ROLLBACK
-- Statement for inserting index of rejected row into temporary table
END CATCH
```

seanfdnn:
`ColumnEncryption` made no difference, unfortunately.

seanfdnn:
[...]

PR #405, opened 2018-06-08
enabling setinputsizes to also work with fast_executemany. 

seanfdnn:
This is a potential fix to Issue #380 that makes the `setinputsizes` behaviour consistent when `fast_executemany=True` 

The logic was copied pretty closely directly from `BindParameter` -- of course ideally, all parameter binding logic should be re-factored out to somewhere common, but that was a little to ambitious for me as a first-time committer.


gordthompson:
This PR includes commit 2f28ec0 which was also included in recently-merged PR #415 so I think this PR can be closed.

mkleehammer:
Closing at gordthompson's suggestion.  Thanks.

Issue #295, opened 2017-10-25
Cursor.fast_executemany with SQL Server temporary table mis-handles Unicode under Python3


gordthompson:
The `Cursor.fast_executemany` feature introduced in 4.0.18/19 does not seem to handle Unicode characters correctly under Python3 when working with a SQL Server temporary table. The test code

```python
import pyodbc
import sys
print('Python version ' + sys.version)
print('pyodbc version ' + pyodbc.version)

conn_str = (
    r'DRIVER=ODBC Driver 11 for SQL Server;'
    r'SERVER=(local)\SQLEXPRESS;'
    r'DATABASE=myDb;'
    r'Trusted_Connection=yes;'
)
cnxn = pyodbc.connect(conn_str, autocommit=True)
crsr = cnxn.cursor()
crsr.fast_executemany = False

crsr.execute("CREATE TABLE #tmp (id INT IDENTITY PRIMARY KEY, txt NVARCHAR(1))")

params = [('Ώ',), ('π',), ('α',)]
crsr.executemany("INSERT INTO #tmp (txt) VALUES (?)", params)

print(crsr.execute("SELECT txt FROM #tmp ORDER BY id").fetchall())
```

works as expected, producing

```
C:\Users\gord\PycharmVirtualenvs\py3pyodbc\Scripts\python.exe C:/Users/gord/PycharmProjects/py3pyodbc_test/main.py
Python version 3.6.2 (v3.6.2:5fd33b5, Jul  8 2017, 04:14:34) [MSC v.1900 32 bit (Intel)]
pyodbc version 4.0.19
[('Ώ', ), ('π', ), ('α', )]

Process finished with exit code 0
```

but when I enable the new feature with `crsr.fast_executemany = True` I get

```
C:\Users\gord\PycharmVirtualenvs\py3pyodbc\Scripts\python.exe C:/Users/gord/PycharmProjects/py3pyodbc_test/main.py
Python version 3.6.2 (v3.6.2:5fd33b5, Jul  8 2017, 04:14:34) [MSC v.1900 32 bit (Intel)]
pyodbc version 4.0.19
[('?', ), ('p', ), ('a', )]

Process finished with exit code 0
```

SQL Profiler shows the following

```
exec sp_describe_undeclared_parameters N'INSERT INTO #tmp (txt) VALUES (@P1)'
exec sp_prepare @p1 output,N'@P1 varchar(255)',N'INSERT INTO #tmp (txt) VALUES (@P1)',1
[...]

PR #415, opened 2018-06-27
Extend setinputsizes for overriding types as well

skeetor:
Refactored the BindParameters and execute_many method to use the same input size override mechanism via a function.
Extended the override method so that it is implemented as intended by the original author to allow overriding of either size or type.

### DIFF
--- a/src/cursor.cpp
+++ b/src/cursor.cpp
@@ -907,9 +907,9 @@ static PyObject* Cursor_executemany(PyObject* self, PyObject* args)
         if (cursor->fastexecmany)
         {
             free_results(cursor, FREE_STATEMENT | KEEP_PREPARED);
-            if (!ExecuteMulti(cursor, pSql, param_seq))
+			if (!ExecuteMulti(cursor, pSql, param_seq))
                 return 0;
-        }
+		}
         else
         {
             for (Py_ssize_t i = 0; i < c; i++)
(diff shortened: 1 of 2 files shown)


gordthompson:
@mkleehammer , @v-chojas , @v-makouz 

I support this change, at least from the functional perspective. (Not qualified to comment on the code changes themselves, sorry.)

There have been quite a few problems stemming from pyodbc trying to deduce what data types are appropriate. It would be very nice to be able to simply *tell* pyodbc exactly what the column types are.

As an added bonus, it allows us to work around the deficiency in SQL Server's `sp_describe_undeclared_parameters` procedure when working with #temporary tables (#295). Under pyodbc 4.0.23 the following test code

```python
import pyodbc
conn_str = (
    r'DRIVER=ODBC Driver 11 for SQL Server;'
    r'SERVER=(local)\SQLEXPRESS;'
    r'DATABASE=myDb;'
    r'Trusted_Connection=yes;'
)
cnxn = pyodbc.connect(conn_str, autocommit=True)
crsr = cnxn.cursor()
crsr.fast_executemany = True

crsr.execute("CREATE TABLE #tmp (id INT IDENTITY PRIMARY KEY, txt NVARCHAR(50))")
sql = "INSERT INTO #tmp (txt) VALUES (?)"
params = [('Ώπα',)]
crsr.setinputsizes([(pyodbc.SQL_WVARCHAR, 50, 0)])
crsr.executemany(sql, params)
print(crsr.execute("SELECT * FROM #tmp").fetchall())
```

prints

```
[(1, '?pa')]
```

but when I run the same code using this build of pyodbc it correctly prints

```
[(1, 'Ώπα')]
```

Issue #371, opened 2018-04-10
Cursor throws an error when trying to insert in SQL Server database when fast_executemany flag is set


manrajgrover:
Please first make sure you have looked at:

* Documentation: https://github.com/mkleehammer/pyodbc/wiki
* Other issues

### Environment

To diagnose, we usually need to know the following, including version numbers.  On Windows, be
sure to specify 32-bit Python or 64-bit:

- Python: 2.7.14
- pyodbc: 4.0.23
- OS: MacOS
- DB: SQL Server
- driver: ODBC Driver 13 for SQL Server

### Issue

Trying to use fast execute many option to insert data to the database throws the error shown below. It works if the flag is not set.

#### Expected behavior:

Cursor should be able to write to database without throwing an error.

#### Observed behavior:

```
---------------------------------------------------------------------------
Error                                     Traceback (most recent call last)
<ipython-input-24-5e5f851229b4> in <module>()
      1 cursor.fast_executemany = True
----> 2 cursor.executemany(query, x.values.tolist())

Error: ('HY010', u'[HY010] [unixODBC][Driver Manager]Function sequence error (0) (SQLParamData)')
```

gizmo93:
The same here:

Python: 3.5.5
pyodbc: 4.0.23
OS: Windows 10
DB: SQL Server 2017 and 2016
driver: ODBC Driver 13 for SQL Server

tried with SQL Server Native Client too. 

Could not determine, where the problem is, but it seems to have to do with string data.

Installed 4.0.22 again and the problems are not there anymore.

v-chojas:
Could you post a example table schema + Python code to reproduce it, and/or ODBC trace log? That will help with the troubleshooting.

gizmo93:
Hey @v-chojas 

It was happening while trying to insert a pandas dataframe with to_sql.

[b64.txt](https://github.com/mkleehammer/pyodbc/files/1922894/b64.txt)

Thats the dataframe, pickled and base64 encoded.
You can use decoded = pickle.loads(base64.b64decode(encoded)) to turn it back into a dataframe and then you can just use encoded.to_sql() to reproduce the error. (Pandas will create a table itself, it it does not exist).

Error Message: Function sequence error (0) (SQLParamData)

gordthompson:
[...]

PR #425, opened 2018-07-16
Fix DAE for 0-length parameters

v-chojas:
Fix for #371


mkleehammer:
Looks interesting.  Would it be easy to add a unit test?  Maybe to SQL Server and PostgreSQL which would allow me to test on Windows, macOS, and Linux?

v-chojas:
I have added tests for SQL Server. SQLDescribeParam in PostgreSQL's ODBC driver doesn't work and always returns a length of 255, so it won't use the DAE path. (The same applies to MySQL.)

ethan-deng:
Still see the issue with 4.0.27 on Windows 10.

Issue #423, opened 2018-07-12
String data, right truncation


srish13:
### Environment
- OS: ubuntu:16.04
- Python: 2.7
- pyodbc:  4.0.23 
- DB:  SQL Server 12
- unixodbc-dev=2.3
-  Microsoft ODBC Driver 13 (13.1.9.2-1)

### Issue
On inserting a string containing emoji characters into an NVARCHAR(MAX) column in MSSQL, I get the following error:
```(pyodbc.DataError) ('22001', u'[22001] [Microsoft][ODBC Driver 13 for SQL Server]String data, right truncation ```.
#### However, it works perfectly fine on windows and mac machine.

On changing python's default encoding to UTF-8, like
```
import sys
reload(sys)
sys.setdefaultencoding('utf-8')
```
I don't get above error. But inserted data are  different symbols.
``` 
sql = "insert into test(user,  category)  values ('abcd', '🌜🌜🌜')" 
cur.execute(sql)
```
ð??¤ð??¤ð??¤ - Inserted in the database. 

```
sql = "insert into test(user, category) values (?, ?)"
cur.execute(sql, ['abcd', '🌜🌜🌜'])
```
If I give parameterized query like above,  ' ???????' are inserted.

- I even tried to change the encoding of pyodbc to (refering this https://github.com/mkleehammer/pyodbc/wiki/Unicode):
1 - con.setencoding(unicode, encoding='utf-8')
2- con.setdecoding(pyodbc.SQL_CHAR, encoding='latin1', to=str)
    con.setencoding(unicode, encoding='raw')
3- cnxn.setdecoding(pyodbc.SQL_CHAR, encoding='latin1', to=str)
cnxn.setencoding(str, encoding='latin1')
 In all the avove cases '??????' is inserted instead of emoticon. 

#### I even tried Microsoft ODBC Driver 17 for SQL Server. But nothing changed.

**Comment by v-chojas** [comment]
What is the exact version of unixODBC? Run `odbcinst -j` to get that information.

Post an [ODBC trace](https://www.simba.com/blog/odbc-troubleshooting-tracing/ ) to help with debugging the problem further.

**Comment by v-chojas** [comment]
Enable tracing in odbcinst.ini and subsequent connections will be traced.

We cannot determine what is happening without an ODBC trace.

**Comment by srish13** [comment]
odbcinst -j output
```
unixODBC 2.3.1
DRIVERS............: /etc/odbcinst.ini
[...]

PR #434, opened 2018-07-30
Fix for truncation error when inserting high unicode characters

v-makouz:


srish13:
@mkleehammer  Because of this issue we are not able to support Unicode characters in our application.  Consider this fix as a high priority as this is blocking developers to use Unicode characters in their application.

mkleehammer:
@v-makouz I realized I've been replying to the issue #423 instead of this PR.  Copying over some of the conversation:

Can we change it to look at the enc.optenc flag instead of enc.name? That way all of the code that looks for the different combinations that mean "utf-16" are in one place (SetTextEncCommon).

Also, can we add some unit tests for the 3 most common.

* PostgreSQL
* SQL Server
* MySQL
* Oracle - I've added this because I think it was mentioned you already had a test.  If not, don't worry about it.  It isn't easy to set up and test.

Here's the test I used for PostgreSQL, but it should work for all of them (it's on the 433-emoticons branch):

    def test_emoticons(self):
        # https://github.com/mkleehammer/pyodbc/issues/423
        #
        # When sending a varchar parameter, pyodbc is supposed to set ColumnSize to the number
        # of characters.  Ensure it works even with 4-byte characters.
        #
        # http://www.fileformat.info/info/unicode/char/1f31c/index.htm

        v = "x \U0001F31C z"

        self.cursor.execute("create table t1(s varchar(100))")
        self.cursor.execute("insert into t1 values (?)", v)

        result = self.cursor.execute("select s from t1").fetchone()[0]

        self.assertEqual(result, v)

mkleehammer:
@v-makouz I pulled and tested and it looks great.  I don't want to steal your authorship and GitHub says it can't automatically rebase, so can you rebase it to a single commit and I'll merge it in?

v-makouz:
@mkleehammer I've made the change to use enc.optenc and I've added the tests, as you said most work as is, but for Python3 version of SQL Server tests I had to replace varchar with nvarchar.

v-makouz:
@mkleehammer I've squished my commits

mkleehammer:
Cherry picked to 686d2b0 since there were other commits between master and it.

Thanks for this fix and the docs pointer.

Closes #434 and I believe #423

srish13:
@mkleehammer When will next release will happen with this change.

mkleehammer:
I will make a release today and include 3.7 binary wheels for Windows and macOS.

PR #447, opened 2018-08-22
tests: remove requirement that driver name be in curly brackets

gordthompson:
It is a common misconception that `DRIVER=` names must be enclosed in curly brackets, e.g.,

```
DRIVER={ODBC Driver 17 for SQL Server}
```

In fact

```
DRIVER=ODBC Driver 17 for SQL Server
```

is perfectly legal. These tweaks allow either form to be used in connection strings for tests.

### DIFF
--- a/tests2/informixtests.py
+++ b/tests2/informixtests.py
@@ -1214,7 +1214,7 @@ class InformixTestCase(unittest.TestCase):
         self.assertEqual(list, type(drivers))
         self.assertTrue(len(drivers) > 1)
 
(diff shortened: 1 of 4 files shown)


v-chojas:
Leave a commentNote that connection string values may themselves contain `}` which must be escaped. This is quite unlikely in the case of those drivers, but just a FYI.

Official documentation: https://sqlprotocoldoc.blob.core.windows.net/productionsqlarchives/MS-ODBCSTR/[MS-ODBCSTR].pdf

mkleehammer:
Thanks @gordthompson.

PR #451, opened 2018-08-25
fix freetdstests errors/failures in tests2

gordthompson:


mkleehammer:
Shouldn't we change the expected type to `unicode` instead of commenting out the lines?

gordthompson:
I just looked at a diff of tests2/freetdstests.py and tests2/sqlservertests.py and they are very similar. It looks like freetdstests.py was forked off sqlservertests.py at commit 83db883 back in November 2011. Since then there have been a number of tests added to sqlservertests.py that haven't made their way into freetdstests.py.

So yes, it probably makes sense to update tests2/sqlservertests.py to also allow testing of FreeTDS ODBC against SQL Server, and retire tests2/freetdstests.py. I can proceed with that if you like.

Issue #480, opened 2018-10-28
Varchar data truncation when fast_executemany is true


keitherskine:
### Environment

- Python: 3.6.5
- pyodbc: 4.0.24
- OS: Windows 10 64-bit
- DB: SQL Server Express (13.0.1742.0)
- driver: ODBC Driver 17 for SQL Server

### Issue

When `fast_executemany` is true, strings that should be too long to be inserted into the table are still inserted into the table, but truncated.

Here is some code that inserts increasingly long strings into a column that is just 5 characters max:

```python
cnxn = pyodbc.connect(MY_CNXN_STR, autocommit=True)
crsr = cnxn.cursor()

sql = "create table dbo.pyodbc_fastexecutemany (id int not null, str1 nvarchar(5) null)"
crsr.execute(sql)

sql = "insert into dbo.pyodbc_fastexecutemany (id, str1) values (?, ?)"

# ids from 1 to 10
data = [(i, 'a' * i) for i in range(1, 11)]
crsr.fast_executemany = False
try:
    crsr.executemany(sql, data)
except Exception as e:
    print(e)

# ids from 1001 to 1010
data = [(i+1000, 'a' * i) for i in range(1, 11)]
crsr.fast_executemany = True
try:
    crsr.executemany(sql, data)
except Exception as e:
    print(e)
```
Here's what ends up in the table:
```
id	str1
1	a
2	aa
3	aaa
4	aaaa
5	aaaaa
1001	a
1002	aa
1003	aaa
1004	aaaa
1005	aaaaa
1006	aaaaa
1007	aaaaa
1008	aaaaa
1009	aaaaa
1010	aaaaa
```
Note the length of the strings for ids 1006 to 1010.

I can kind of understand that when `fast_executemany` is false and `autocommit` is true, the first 5 records would be inserted into the table and the sixth would error out.  However, what seems very odd is that when `fast_executemany` is true, strings which are too long for the column are apparently truncated before inserting them anyway.  And this happens even though it appears some kind of error is returned by the database.  I would have thought the whole transaction should be rolled back, so that the end result is no records with ids 1000+ at all would end up in the table.  At the very least, string values should not be truncated.

Also, when `fast_executemany` is false, this error is printed out:
```
[...]

PR #482, opened 2018-10-30
Fix propagation of truncation errors

v-makouz:
a


v-chojas:
For fix  #480

v-chojas:
"String data, right truncation" is a standard ODBC error message.

https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqlexecute-function

> The assignment of a character or binary value to a column resulted in the truncation of nonblank (character) or non-null (binary) characters or bytes.

keitherskine:
That's good to know, @v-chojas , many thanks.

PR #483, opened 2018-11-03
fix cursor doc strings for fetchall and fetchmany

rosscoleman:
I noticed that the docstrings were mixed up for cursor.fetchall and cursor.fetchmany.


rosscoleman:
I noticed that the docstrings were mixed up for cursor.fetchall and cusor.fetchmany.

Issue #496, opened 2018-12-19
Retrieve currently active Output Converter function(s)


gordthompson:
With the addition of `Connection#remove_output_converter` in 4.0.25 it is easy to temporarily invoke an Output Converter function for a special case, e.g.,

```python
# workaround driver bug
cnxn.add_output_converter(pyodbc.SQL_WVARCHAR, decode_sketchy_utf16)
rows = crsr.columns("Clients").fetchall()
cnxn.remove_output_converter(pyodbc.SQL_WVARCHAR)  # restore default behaviour
```

A nice enhancement would be a method to retrieve the current Output Converter function (if any) so we can restore the previous behaviour.

```python
# retrieve current converter
prev_converter = cnxn.get_output_converter(pyodbc.SQL_WVARCHAR)  # new Connection method
print(type(prev_converter))  # <class 'function'>
print(repr(prev_converter))  # <function our_usual_custom_utf16_decoder at 0x00000000029DB840>

# workaround driver bug
cnxn.add_output_converter(pyodbc.SQL_WVARCHAR, decode_sketchy_utf16)  # replace current converter
rows = crsr.columns("Clients").fetchall()

# restore previous behaviour
cnxn.add_output_converter(pyodbc.SQL_WVARCHAR, prev_converter)
```

PR #497, opened 2018-12-20
add get_output_converter method (#496)

gordthompson:
Add get_output_converter method to Connection object as described in #496


mkleehammer:
Thanks so much @gordthompson .  I love seeing the tests in there too.  Sorry for the delay.

Issue #509, opened 2019-01-16
4.0.25 breaks queries with integer parameters against Access database


gordthompson:
(Prompted by [this Stack Overflow question](https://stackoverflow.com/q/54220640/2144390).)

- Python version 3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 23:09:28) [MSC v.1916 64 bit (AMD64)]
- pyodbc version 4.0.25
- driver version ACEODBC.DLL  Microsoft Access database engine

A change introduced in 4.0.25 seems to have broken queries with integer parameters against an Access database.

#### 4.0.25:

```python
existquery = """SELECT * FROM EconVars WHERE Year(ValueDate) = ?"""
params = (1999,)
crsr.execute(existquery,params)
```

produces

```
Traceback (most recent call last):
  File "C:/Users/Gord/PycharmProjects/py3pyodbc_demo/main.py", line 106, in <module>
    crsr.execute(existquery,params)
pyodbc.Error: ('HYC00', '[HYC00] [Microsoft][ODBC Microsoft Access Driver]Optional feature not implemented  (106) (SQLBindParameter)')
```

specifically:

```
main            154c-2284   EXIT  SQLBindParameter  with return code -1 (SQL_ERROR)
        HSTMT               0x000000000021EDB0
        UWORD                        1 
        SWORD                        1 <SQL_PARAM_INPUT>
        SWORD                      -25 <SQL_C_SBIGINT>
        SWORD                       -5 <SQL_BIGINT>
        SQLULEN                    0
        SWORD                        0 
        PTR                0x0000000002C46A98
        SQLLEN                     0
        SQLLEN *            0x0000000002C46A68

        DIAG [HYC00] [Microsoft][ODBC Microsoft Access Driver]Optional feature not implemented  (106) 
```

However, this does work:

```python
existquery = """SELECT * FROM EconVars WHERE Str(Year(ValueDate)) = ?"""
params = (' 1999',)
crsr.execute(existquery,params)
```

#### 4.0.24:

The original query works. The ODBC log shows

```
main            d1c-1570    EXIT  SQLBindParameter  with return code 0 (SQL_SUCCESS)
        HSTMT               0x02AA1458
        UWORD                        1 
        SWORD                        1 <SQL_PARAM_INPUT>
        SWORD                        4 <SQL_C_LONG>
        SWORD                        4 <SQL_INTEGER>
        SQLULEN                    0
        SWORD                        0 
[...]

PR #529, opened 2019-02-25
Use int param instead of bigint when possible

mkleehammer:
Fix for #509.  Access int params no longer work because we always use bigint in Python 3.  Changing to choose int or bigint based on the actual value.

@v-makouz Can you make sure your TVP code still works?  It kept accessing PyError_Occurred and I'm not sure why.  I don't see why all of the binding code wouldn't be the same.  Are you sure there wasn't some other bug in the TVP code that was being masked or are we possibly missing errors in non-int code?

Also, sorry for the massive number of whitespace changes.  I didn't notice until I created the pull request.  Please commit with spaces and no trailing whitespace from now on.


mkleehammer:
This is a work in progress - I'm getting lots of SQL Server errors.  Once I'm finished I'll rebase to a clean commit.  I just wanted everyone to see my proposed fix as I work through it.

v-makouz:
@mkleehammer I just reran the sqlservertests.py on Windows and Linux using python 2.7 and 3.6 and they appear to still be passing with the latest version of pyODBC

v-makouz:
@mkleehammer On the `509-access-ints` branch, I do get some failures: test_bigint on Linux and a 10 different ones on Windows, but none are in TVP tests, those still seem to be OK

gordthompson:
I just built `509-access-ints` (d9e90fd) under 64-bit Python 3.7.2 on Windows_7 and ran `tests3\accesstests.py`:

```
C:\Users\Gord\git\pyodbc>python3 tests3\accesstests.py \Users\Public\Database1.accdb
DRIVER={Microsoft Access Driver (*.mdb, *.accdb)};DBQ=C:\Users\Public\Database1.accdb;ExtendedAnsiSQL=1
python:  3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 23:09:28) [MSC v.1916 64 bit (AMD64)]
pyodbc:  4.0.0-unsupported C:\Users\Gord\git\pyodbc\build\lib.win-amd64-3.7\pyodbc.cp37-win_amd64.pyd
odbc:    03.80.0000
driver:  ACEODBC.DLL Microsoft Access database engine
         supports ODBC version 03.51
os:      Windows
unicode: Py_Unicode=2 SQLWCHAR=2
Max VARCHAR = (not supported)
Max WVARCHAR = 255
Max BINARY = 510
         7 6.1.7601 SP1 Multiprocessor Free
FAIL: test_executemany (__main__.AccessTestCase)
  File "tests3\accesstests.py", line 516, in test_executemany
    self.assertEqual(param[0], row[0])
AssertionError: 1 != 0

FAIL: test_int (__main__.AccessTestCase)
  File "tests3\accesstests.py", line 294, in test_int
    self.assertEqual(result, value)
AssertionError: 0 != 1234

FAIL: test_negative_int (__main__.AccessTestCase)
  File "tests3\accesstests.py", line 301, in test_negative_int
    self.assertEqual(result, value)
AssertionError: 0 != -1

FAIL: test_smallint (__main__.AccessTestCase)
  File "tests3\accesstests.py", line 308, in test_smallint
    self.assertEqual(result, value)
AssertionError: 0 != 32767

FAIL: test_subquery_params (__main__.AccessTestCase)
Ensure parameter markers work in a subquery
  File "tests3\accesstests.py", line 239, in test_subquery_params
    self.assertEqual(row[0], 1)
AssertionError: 0 != 1

FAIL: test_tinyint (__main__.AccessTestCase)
  File "tests3\accesstests.py", line 344, in test_tinyint
    self.assertEqual(value, result)
AssertionError: 10 != 0

Ran 72 tests in 5.304s

FAILED (failures=6)
```

mkleehammer:
I know everything is broken - I'm still working on it.  I made the initial changes on macOS and needed a branch to test them on Windows.  Now working through.  SQL Server is working again with new code (with 2nd commit) in 32-bit and 64-bit.  Now working on Access again.

mkleehammer:
I updated the access tests to accept the extension only and to copy the empty.accdb or empty.mdb to a test file on its own.  The test.accdb and test.mdb files are now ignored.  That was always a pain.  It isn't perfect since you still can't run the new driver with an MDB file so we really need to pass two command line options.  (Note that this is not a new limitation.)

All tests pass in 64-bit Python 3.6 but not 2.7.  Still looking.

mkleehammer:
@gordthompson Try now.  I set the Python 2 unit tests to encode the `str` type (so `bytes`, basically) as UTF-16 due to this article: [Desktop Database Driver Performance Issues](https://docs.microsoft.com/en-us/sql/odbc/microsoft/desktop-database-driver-performance-issues?view=sql-server-2017).  I'm not sure if that is right but it works.

I tested with the driver I had which is the "accdb" one: ACEODBC.DLL Microsoft Access database engine 14.00.7180.5000.  It is the 64-bit version.  I need to find the 32-bit version to test it.

mkleehammer:
To clarify how to run the unit tests now:

    python accesstests.py -v accdb

gordthompson:
Just tested d9e90fd on 32-bit Win_7 with Python 2.7 and 3.6. All accesstests.py tests pass for tests2\ and tests3\

PR #532, opened 2019-02-26
tweak test_emoticons in mysqltests.py (#536)

gordthompson:


mkleehammer:
Thanks @gordthompson.  I know anything from you is going to be good.

PR #564, opened 2019-04-26
Fix UTF-16 conversion issues for connection errors on big endian

### LINKED ISSUE #563: Connection errors give wrong message string on big-endian systems
### Environment

- Python: 3.6.8 64-bit
- pyodbc: 4.0.27b1
- OS: IBM i 7.2
- DB: N/A
- driver: N/A

### Issue

Connection errors gives mojibake on big-endian systems:

```
>>> pyodbc.connect('DSN=doesnotexist')
pyodbc.InterfaceError: ('IM002', '[IM002] 嬀甀渀椀砀伀䐀䈀䌀崀嬀䐀爀椀瘀攀爀\u2000䴀愀渀愀最攀爀崀䐀愀琀愀\u2000猀漀甀爀挀攀\u2000渀愀洀攀\u2000渀漀琀\u2000昀漀甀渀搀Ⰰ\u2000愀渀搀\u2000渀漀\u2000搀攀昀愀甀氀琀\u2000搀爀椀瘀攀爀\u2000猀瀀攀挀椀昀椀攀搀 (0) (SQLDriverConnect)')
```

This is due to hardcoding the encoding to "utf-16-le" when a connection is NULL: https://github.com/mkleehammer/pyodbc/blob/master/src/errors.cpp#L294

If I change the encoding to "utf-16-be", I get the correct error message:

```
>>> pyodbc.connect('DSN=doesnotexist')
pyodbc.InterfaceError: ('IM002', '[IM002] [unixODBC][Driver Manager]Data source name not found, and no default driver specified (0) (SQLDriverConnect)')
```

Looks like this should use `ENCSTR_UTF16NE` which seems to have been missed from #255.

kadler:
In #255, the code was changed to use the native endianness of the
system, but when a connection failure occurs the encoding was still
hard-coded to utf-16-le which causes mojibake on big endian systems.

Instead use ENCSTR_UTF16NE, which will be either utf-16-le or
utf-16-be, depending on the native endian for the system.

Fixes #563

### DIFF
--- a/src/errors.cpp
+++ b/src/errors.cpp
@@ -289,9 +289,9 @@ PyObject* GetErrorFromHandle(Connection *conn, const char* szFunction, HDBC hdbc
         // Not always NULL terminated (MS Access)
         sqlstateT[5] = 0;
 
-        // For now, default to UTF-16LE if this is not in the context of a connection.
+        // For now, default to UTF-16 if this is not in the context of a connection.
         // Note that this will not work if the DM is using a different wide encoding (e.g. UTF-32).
-        const char *unicode_enc = conn ? conn->metadata_enc.name : "utf-16-le";
+        const char *unicode_enc = conn ? conn->metadata_enc.name : ENCSTR_UTF16NE;
         Object msgStr(PyUnicode_Decode((char*)szMsg, cchMsg * sizeof(ODBCCHAR), unicode_enc, "strict"));
 
         if (cchMsg != 0 && msgStr.Get())


jkyeung:
<blockquote>
system, but when a connection failure occurs the encoding was still
hard-coded to utf-16-le which causes mojibake on big endian systems.
</blockquote>

Seems to be missing some words there. Not that it matters much; the issue and the fix are both very simple to understand.

kadler:
Oops, I referenced #255 in the git commit message, but git automatically eats lines that start with #

mkleehammer:
Great catch.  Thanks.

Issue #596, opened 2019-07-27
Type mismatch for None/NULL value in a cell in a table-valued parameter.


mhmurray:
# Sys info
- python:  3.6.1 |Anaconda custom (64-bit)| (default, May 11 2017, 13:09:58)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]
- pyodbc:  4.0.26 [...]/lib/python3.6/site-packages/pyodbc.cpython-36m-x86_64-linux-gnu.so
- odbc:    03.52

- driver:  libmsodbcsql-17.3.so.1.1 17.03.0001
         supports ODBC version 03.52
- os:      Ubuntu 16.04
- DB: Azure SQL Server

# Issue
When using a table-valued parameter (TVP) for a simple stored procedure that echos the input, `None` values in the TVP cells cause a pyodbc error because of a type mismatch with the parameter in the previous row.

Stored procedure:
`CREATE PROCEDURE TestSelectTVP @TVP TVPType READONLY AS SELECT * FROM @TVP;`

TVP Type:
`CREATE TYPE TVPType AS TABLE(c01 BIGINT, c02 BIGINT)`

Call the SP:
```
param_array = [
        [1, 0],
        [None, 1],
        ]
cursor.execute("EXEC TestSelectTVP ?", [param_array])
```

## Expected
Returned values:
```
c0   | c1
-----|-----
1    | 0
NULL | 1
```

## Observed
Exception raised:
```
Traceback (most recent call last):
  File "test_pyodbc.py", line 144, in <module>
    main()
  File "test_pyodbc.py", line 139, in main
    run_tvp(conn)
  File "test_pyodbc.py", line 103, in run_tvp
    cursor.execute("EXEC TestSelectTVP ?", [param_array])
pyodbc.ProgrammingError: ('Type mismatch between TVP row values', 'HY000')
```
If None/NULL is in the first row, there is a different exception:
`pyodbc.Error: ('HY090', '[HY090] [unixODBC][Driver Manager]Invalid string or buffer length (0) (SQLBindParameter)')`

## Other info

The exception is raised here:
https://github.com/mkleehammer/pyodbc/blob/4.0.26/src/cursor.cpp#L766

Doesn't this type check require non-NULL data, since it's testing strict type equality?

# Script to reproduce:
```
import pyodbc
import sys
import os
import platform


CONNECTION_STR=''

def reset_tvp(conn):
    """
    Reset stored procedures and table types associated with the TVP method.
[...]

PR #605, opened 2019-08-09
Fix for 596

v-makouz:
https://github.com/mkleehammer/pyodbc/issues/596

I tried this with the repro app, and ran the sqlserver tests, everything seems to work OK


Zeddicus414ce:
I ran into #596 as well. I was glad to see that this is in work! Thank you.

I was using the test_tvp code to track down the problem before I found this ticket. Would adding a None/NULL input to the test_tvp be helpful here? Perhaps changing c05 to [1, 0, None]?

eydelrivero:
Glad to see this has been fixed. Will this pull request be merged any time soon?

githubuserkkk:
I still see the issue if we have None in the first row. ('HY004', '[HY004] [Microsoft][SQL Server Native Client 11.0]Invalid SQL data type (0) (SQLBindParameter)

vikas-clinch:
I am also

> I still see the issue if we have None in the first row. ('HY004', '[HY004] [Microsoft][SQL Server Native Client 11.0]Invalid SQL data type (0) (SQLBindParameter)

Me too

himanshu754:
> I still see the issue if we have None in the first row. ('HY004', '[HY004] [Microsoft][SQL Server Native Client 11.0]Invalid SQL data type (0) (SQLBindParameter)

I am also facing this issue

PR #611, opened 2019-08-20
Bind 32-bit ints correctly on big endian LP64 systems

kadler:
The size of a long on 64-bit Linux, Unix, and other LP64 platforms
is 64-bit, so passing the address of it and saying it is a 32-bit
integer will result in only the top 4 bytes being read. While this
does work on little endian systems, on big endian systems these bytes
will always be 0 for values < INT_MAX.

Instead, change the type to be an int, which should be 32 bits on
all supported platforms.

### DIFF
--- a/src/cursor.h
+++ b/src/cursor.h
@@ -72,7 +72,7 @@ struct ParamInfo
     union
     {
         unsigned char ch;
-        long l;
+        int i32;
         INT64 i64;
         double dbl;
         TIMESTAMP_STRUCT timestamp;
--- a/src/params.cpp
+++ b/src/params.cpp
@@ -880,11 +880,11 @@ static bool GetIntInfo(Cursor* cur, Py_ssize_t index, PyObject* param, ParamInfo
     }
     else
     {
-        info.Data.l = PyLong_AsLong(param);
+        info.Data.i32 = PyLong_AsLong(param);
 
         info.ValueType         = SQL_C_LONG;
         info.ParameterType     = SQL_INTEGER;
-        info.ParameterValuePtr = &info.Data.l;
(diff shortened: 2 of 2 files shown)


mkleehammer:
I see what you mean.  For decades MS has ended up with defines for the wrong sizes (WPARAM anyone?) because they designed them for the current environment.

However, shouldn't we try to use `int32` (or INT32 if we have to define it) instead of `int`?  The natural word size, which `int` is defined to be, would be 64-bits on these environments, wouldn't it?

v-chojas:
`int` is still 32-bits with the LP64 data model:

https://en.wikipedia.org/wiki/LP64#64-bit_data_models

mkleehammer:
You are correct - I was thinking Mac was ILP.  It should have occurred to me when the current code gave me compiler warnings for assigning `PyLong_AsLong` to an int.

How about just eliminating the new warnings by casting the result to `(int)` then?  I'll then squash and merge.

Also, does anyone know a simple workflow for collectively working on a PR?  I assume I can't commit to this branch (and it might be rude to push to someone else's branch).  However, there must be some way for me to commit a suggestion based off of someone else's PR code.  Should I just make a branch based on it and publish it?

kadler:
>  Also,  does anyone know a simple workflow for collectively working on a PR?  I  assume I can't commit to this branch (and it might be rude to push to  someone else's branch).  However, there must be some way for me to  commit a suggestion based off of someone else's PR code.  Should I just  make a branch based on it and publish it?

FYI, GitHub has an option to allow maintainers to make changes to a PR: https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

I don't recall if I enabled it on this PR or not, though.

PR #621, opened 2019-09-16
Expose SQL Types in Cursor

celestialorb:
This is a follow-up PR for issue #167, which has had a PR submitted already (#168). When testing this modified `pyodbc` package locally I found that the `coldescription` attribute of the cursor was not set properly for procedures that return more than one result set.

### DIFF
--- a/src/cursor.cpp
+++ b/src/cursor.cpp
@@ -1883,6 +1883,9 @@ static PyObject* Cursor_nextset(PyObject* self, PyObject* args)
 
         if (!create_name_map(cur, cCols, lowercase()))
             return 0;
+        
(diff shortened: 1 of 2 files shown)


mkleehammer:
Very nice and clean.

I know it seems analogous to `Cursor.description`, but I am concerned about adding any more processing for something that will be rarely used.  When needed it is important, but I think it won't be needed most of the time.  What do you think about turning it into a function instead?

It might also be nice to figure out if we can easily make the return type a  list of `namedtuple`s.  Then we could document that attributes may be added to the tuple in the future, so use the names instead of tuple unpacking.

mkleehammer:
I've put this on hold for v5 due to the suggestions I made above.

PR #623, opened 2019-09-17
Fix for issue 617

v-makouz:
Fixing the insertion of high unicode characters, bug caused by SQLPutData being given the number of characters, rather then code points, which causes problems for characters are more code points then usual


mkleehammer:
Looks good.  Small change request - could you change the table to "t1".  Tests use t0-t2 which are automatically dropped before each test.  (Also, I'm not sure why it is using `executemany` instead of `execute`.  Is that on purpose?)

v-makouz:
I renamed the table, as for why it's `executemany` I simply based the test on the OP's repro script, I wanted to keep it as close as possible, since that's what originally indicated the problem

PR #637, opened 2019-10-21
Add emoticon-as-literal tests for PostgreSQL.

gordthompson:
As with MySQL Connector/ODBC, the PostgreSQL ODBC driver has both "ANSI" and "Unicode" versions. Make sure that at least one of them ("Unicode", in this case) can handle Unicode literals with characters outside of the BMP.

### DIFF
--- a/tests2/pgtests.py
+++ b/tests2/pgtests.py
@@ -16,6 +16,7 @@ You can also put the connection string into a tmp/setup.cfg file like so:
   [pgtests]
   connection-string=DSN=PostgreSQL35W
 
+Note: Be sure to use the "Unicode" (not the "ANSI") version of the PostgreSQL ODBC driver.
 """
 
 import sys, os, re
@@ -496,7 +497,7 @@ class PGTestCase(unittest.TestCase):
             v = self.cursor.execute("select a from t1").fetchone()[0]
             self.assertEqual(v, value)
             
-    def test_emoticons(self):
+    def test_emoticons_as_parameter(self):
         # https://github.com/mkleehammer/pyodbc/issues/423
         #
         # When sending a varchar parameter, pyodbc is supposed to set ColumnSize to the number
@@ -506,8 +507,20 @@ class PGTestCase(unittest.TestCase):
 
         v = "x \U0001F31C z"
 
-        self.cursor.execute("create table t1(s varchar(100))")
+        self.cursor.execute("CREATE TABLE t1(s varchar(100))")
         self.cursor.execute("insert into t1 values (?)", v)
+
+        result = self.cursor.execute("select s from t1").fetchone()[0]
+
+        self.assertEqual(result, v)
+
+    def test_emoticons_as_literal(self):
+        # https://github.com/mkleehammer/pyodbc/issues/630
+
+        v = "x \U0001F31C z"
+
+        self.cursor.execute("CREATE TABLE t1(s varchar(100))")
+        self.cursor.execute("insert into t1 values ('%s')" % v)
 
         result = self.cursor.execute("select s from t1").fetchone()[0]
 
--- a/tests3/pgtests.py
+++ b/tests3/pgtests.py
@@ -16,6 +16,7 @@ You can also put the connection string into a tmp/setup.cfg file like so:
   [pgtests]
   connection-string=DSN=PostgreSQL35W
 
+Note: Be sure to use the "Unicode" (not the "ANSI") version of the PostgreSQL ODBC driver.
 """
 
 import uuid
@@ -593,7 +594,7 @@ class PGTestCase(unittest.TestCase):
         self.cursor.execute("select 1")
         self.cursor.cancel()
 
-    def test_emoticons(self):
+    def test_emoticons_as_parameter(self):
         # https://github.com/mkleehammer/pyodbc/issues/423
         #
         # When sending a varchar parameter, pyodbc is supposed to set ColumnSize to the number
@@ -603,8 +604,20 @@ class PGTestCase(unittest.TestCase):
 
         v = "x \U0001F31C z"
 
-        self.cursor.execute("create table t1(s varchar(100))")
+        self.cursor.execute("CREATE TABLE t1(s varchar(100))")
         self.cursor.execute("insert into t1 values (?)", v)
+
+        result = self.cursor.execute("select s from t1").fetchone()[0]
+
+        self.assertEqual(result, v)
+
+    def test_emoticons_as_literal(self):
+        # https://github.com/mkleehammer/pyodbc/issues/630
+
+        v = "x \U0001F31C z"
+
+        self.cursor.execute("CREATE TABLE t1(s varchar(100))")
+        self.cursor.execute("insert into t1 values ('%s')" % v)
 
         result = self.cursor.execute("select s from t1").fetchone()[0]
 


keitherskine:
Hi @gordthompson , could I ask, did all the PostgreSQL unit tests pass after adding this test?  Specifically, did they pass when using Python 3 on Ubuntu 16.04?  All the tests pass on Windows, and pass when using Python 2.7 on Ubuntu 16.04, but with Python 3 on Ubuntu 16.04, all the unicode tests fail:

(some of the unicode characters don't seem to have copy-and-pasted very well here)
```
$ python "./$TESTS_DIR/pgtests.py" "DRIVER={PostgreSQL Unicode};SERVER=localhost;UID=postgres;DATABASE=test"
ERROR: test_columns (__main__.PGTestCase)
  File "./tests3/pgtests.py", line 576, in test_columns
    row = results['xΏz']
KeyError: 'xΏz'
FAIL: test_chinese (__main__.PGTestCase)
  File "./tests3/pgtests.py", line 215, in test_chinese
    self.assertEqual(row[0], v)
AssertionError: 'æ\x88\x91ç\x9a\x84' != '我的'
- æç
+ 我的
FAIL: test_emoticons_as_literal (__main__.PGTestCase)
  File "./tests3/pgtests.py", line 625, in test_emoticons_as_literal
    self.assertEqual(result, v)
AssertionError: 'x ð\x9f\x8c\x9c z' != 'x 🌜 z'
- x ð z
+ x 🌜 z
Ran 78 tests in 0.592s
FAILED (failures=2, errors=1)
```

gordthompson:
> 
> 
> Hi @gordthompson , could I ask, did all the PostgreSQL unit tests pass after adding this test? Specifically, did they pass when using Python 3 on Ubuntu 16.04?

Hi Keith. TBH, I don't specifically remember which platform(s) I tried. Very likely Windows, and possibly some version of Ubuntu (depending on which VMs I had available at the time). Certainly if you are getting consistent failures under Python 3 on Linux we should investigate further. (Hey, that's what tests are for, right? ☺️ )

keitherskine:
OK, good to know, Gord.  I'll see if I can figure out what's going on.  I get the same failures with Python 3.6, 3.7, and 3.8.

gordthompson:
Here's a test on an install of Xubuntu 18.04 that has been upgraded from the Ubuntu repo's unixODBC 2.3.4 to the latest 2.3.7. ([This page](http://www.unixodbc.org/) says that 2.3.5 contained "Many bug fixes".)

```text
gord@gord-dv7-xubuntu0:~/git/pyodbc$ python3 tests3/pgtests.py -v
/home/gord/git/pyodbc/tests3 --> /home/gord/git/pyodbc
python:  3.6.9 (default, Apr 18 2020, 01:56:04) 
[GCC 8.4.0]
pyodbc:  4.0.30 /home/gord/git/pyodbc/build/lib.linux-x86_64-3.6/pyodbc.cpython-36m-x86_64-linux-gnu.so
odbc:    03.52
driver:  psqlodbcw.so 12.01.0000
         supports ODBC version 03.51
os:      Linux
unicode: Py_Unicode=4 SQLWCHAR=2
Max VARCHAR = 255
Max WVARCHAR = 255
Max BINARY = (not supported)
......................F.......................................................
FAIL: test_emoticons_as_literal (__main__.PGTestCase)
  File "tests3/pgtests.py", line 625, in test_emoticons_as_literal
    self.assertEqual(result, v)
AssertionError: 'x ð\x9f\x8c\x9c z' != 'x 🌜 z'
- x ð z
+ x 🌜 z


Ran 78 tests in 4.240s

FAILED (failures=1)
```

The "funny character" (🌜) is U+1F31C which is UTF-8 encoded as `F0 9F 8C 9C`.

Hope this helps.

gordthompson:
... and here's just the one test ...

```text
gord@gord-dv7-xubuntu0:~/git/pyodbc$ python3 tests3/pgtests.py -t test_emoticons_as_literal -v
/home/gord/git/pyodbc/tests3 --> /home/gord/git/pyodbc
python:  3.6.9 (default, Apr 18 2020, 01:56:04) 
[GCC 8.4.0]
pyodbc:  4.0.30 /home/gord/git/pyodbc/build/lib.linux-x86_64-3.6/pyodbc.cpython-36m-x86_64-linux-gnu.so
odbc:    03.52
driver:  psqlodbcw.so 12.01.0000
         supports ODBC version 03.51
os:      Linux
unicode: Py_Unicode=4 SQLWCHAR=2
Max VARCHAR = 255
Max WVARCHAR = 255
Max BINARY = (not supported)
F
FAIL: test_emoticons_as_literal (__main__.PGTestCase)
  File "tests3/pgtests.py", line 625, in test_emoticons_as_literal
    self.assertEqual(result, v)
AssertionError: 'x ð\x9f\x8c\x9c z' != 'x 🌜 z'
- x ð z
+ x 🌜 z


Ran 1 test in 0.532s

FAILED (failures=1)
```

... and the ODBC trace ...

[odbctrace.txt](https://github.com/mkleehammer/pyodbc/files/4691355/odbctrace.txt)

... in case you want to open an issue.

gordthompson:
This might be relevant

```text
[ODBC][1493][1590608274.927362][SQLDriverConnectW.c][290]
        Entry:
            Connection = 0x1d9c980
            Window Hdl = (nil)
            Str In = [DSN=pyodbc_test_Unicode;UID=postgres;PWD=_what_ever_][length = 51 (SQL_NTS)]
            Str Out = (nil)
            Str Out Max = 0
            Str Out Ptr = (nil)
            Completion = 0
        UNICODE Using encoding ASCII 'UTF-8' and UNICODE 'UCS-2LE'
```

because I'm pretty sure that UCS-2 can't do emoji.

v-chojas:
You can try to rebuild unixODBC with configure option `--with-iconv-ucode-enc=UTF-16LE`.

keitherskine:
Thanks for all those updates, Gord.  Strange you get only one error whereas I get three.

I tried rebuilding unixODBC with `--with-iconv-ucode-enc=UTF16LE` and `--with-iconv-char-enc=UTF8` (ref [here](https://docs.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-driver-manager?view=sql-server-ver15)), but I still get three errors.  Hmmm.

Btw, I updated my original comment with the verbose header info.  All seems similar to Gord's setup.

UCS-2 is not _supposed_ to be able to do emojis (i.e. code points beyond the basic plane), but it seems to morph into UTF16 when the need arises.  For example, all these emoji tests work on Windows with SQL Server, which is supposed to be the most strict adherent to UCS-2.  In than case, emojis get passed back as 4-byte characters, as if it were UTF16.

gordthompson:
I suppose one other possible difference might be the server version. I'm hitting a version 12.3 server installed from the PostgreSQL repositories at `apt.postgresql.org`.

keitherskine:
Hmm, if I use Ubuntu 18.04 with PostgreSQL 10 or 11, then I get just the one error.  On Ubuntu 16.04 with PostgreSQL 9 or 10, I was getting three errors.  Some progress I suppose.  Don't know what the crucial difference is between Ubuntu 16 and 18, but it seems to have helped.  Btw, in all cases I upgraded unixODBC to v2.3.7.  Does the order matter here though?  If PostgreSQL is started up before unixODBC is upgraded, does PostgreSQL ignore the change somehow?

keitherskine:
This seems to be a write issue with high unicode characters, and I'm pretty sure it's a driver issue, not pyodbc.  If you write high unicode characters using PostgreSQL [unicode notation](https://www.postgresql.org/docs/11/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-UESCAPE), the string is read back just fine. But if you embed the high unicode character within the SQL string, it's mangled when it's read back.  For example:

```python
# PostgreSQL
crsr.execute("CREATE TABLE t1(id int, s1 varchar(100), s2 varchar(100))")
v1 = "x \U0001F31C z"  # high unicode, half moon emoji 🌜
v2 = "\u6211\u7684"    # normal (2-byte) unicode, chinese characters 我的
crsr.execute(r"insert into t1 values (1, U&'x \+01F31C z', '{}')".format(v1))
crsr.execute(r"insert into t1 values (2, U&'\6211\7684', '{}')".format(v2))
```
Both the normal unicode chinese characters are fine when read from the database.  For the high unicode emoji character, column s1 (postgres unicode notation) is read back just fine, but column s2 (embedded character) is gibberish.

```
(writing)
[ODBC][25398][1590664490.057932][SQLExecDirect.c][240]
		Entry:
			Statement = 0x5612f0515100
			SQL = [insert into t1 values (1, U&'x \+01F31C z', 'x 🌜 z')][length = 55]
[ODBC][25398][1590664490.058301][SQLExecDirect.c][515]
		Exit:[SQL_SUCCESS]

...

(reading)
[ODBC][25398][1590664490.061578][SQLGetData.c][237]
		Entry:
			Statement = 0x5612f0515100
			Column Number = 2
			Target Type = 1 SQL_CHAR
			Buffer Length = 4096
			Target Value = 0x5612f051bfb0
			StrLen Or Ind = 0x7ffcfd40cfa8
[ODBC][25398][1590664490.061584][SQLGetData.c][534]
		Exit:[SQL_SUCCESS]                
			Buffer = [x 🌜 z]                
			Strlen Or Ind = 0x7ffcfd40cfa8 -> 8
[ODBC][25398][1590664490.061593][SQLGetData.c][237]
		Entry:
			Statement = 0x5612f0515100
			Column Number = 3
			Target Type = 1 SQL_CHAR
			Buffer Length = 4096
			Target Value = 0x5612f051bfb0
			StrLen Or Ind = 0x7ffcfd40cfa8
[ODBC][25398][1590664490.061599][SQLGetData.c][534]
		Exit:[SQL_SUCCESS]                
			Buffer = [x ð z]                
			Strlen Or Ind = 0x7ffcfd40cfa8 -> 12
```
So, what should we do about that PostgreSQL "test_emoticons_as_literal" unit test?  I know it passes on Windows, but should we comment it out for now?  Or run it only on Windows with a skip decorator?
```
    @unittest.skipIf(sys.platform != 'win32', "high unicode embedded characters not supported yet")
    def test_emoticons_as_literal(self):
```

gordthompson:
I'd vote for "skip if not Windows", or something like I did for the SQL Server tests and FreeTDS ODBC. If we can clearly demonstrate that pyodbc is sending the literal to the driver correctly, then we should also open a ticket for PostgreSQL ODBC and put the ticket # in the comments for the test.

gordthompson:
Even more curiouser, this super basic round-trip test passes on the same Linux VM as my tests above:

```python
import os
import sys

import pyodbc

driver = (
    r"C:\Program Files (x86)\psqlODBC\1200\bin\psqlodbc35w.dll"
    if sys.platform == "win32"
    else "/usr/lib/x86_64-linux-gnu/odbc/psqlodbcw.so"
)
connection_string = (
    f"Driver={driver};"
    "CommLog=0;"
    "Debug=2;"
    "Fetch=100;"
    "UniqueIndex=1;"
    "UseDeclareFetch=0;"
    "UnknownSizes=0;"
    "TextAsLongVarchar=1;"
    "UnknownsAsLongVarchar=0;"
    "BoolsAsChar=1;"
    "Parse=0;"
    "MaxVarcharSize=255;"
    "MaxLongVarcharSize=8190;"
    "ExtraSysTablePrefixes=;"
    "Description=;"
    "Database=postgres;"
    "Servername=192.168.0.199;"
    "Port=5432;"
    "Username=postgres;"
    "UID=postgres;"
    f"Password={os.getenv('my_PWD')};"
    "ReadOnly=0;"
    "ShowOidColumn=0;"
    "FakeOidIndex=0;"
    "RowVersioning=0;"
    "ShowSystemTables=0;"
    "Protocol=;"
    "ConnSettings=;"
    "pqopt=;"
    "UpdatableCursors=1;"
    "LFConversion=1;"
    "TrueIsMinus1=0;"
    "BI=0;"
    "D6=-101;"
    "AB=0;"
    "ByteaAsLongVarBinary=1;"
    "UseServerSidePrepare=1;"
    "LowerCaseIdentifier=0;"
    "SSLmode=disable;"
    "KeepaliveTime=-1;"
    "KeepaliveInterval=-1;"
    "XaOpt=1;"
)
test_string = 'x 🌜 y'
cnxn = pyodbc.connect(connection_string)
result = cnxn.execute(f"SELECT '{test_string}' AS foo").fetchval()
if result == test_string:
    print("pass")
else:
    print(f"FAIL: result={repr(result)} ; expected={repr(test_string)}")
```

gordthompson:
Notes on my last post:

1. I got those connection string attributes by using [dump_dsn](https://github.com/gordthompson/dump_dsn) on the DSN that was generated by the ODBC Administrator on Windows. My original intent was just to provide an explicit self-contained example, but then it occurred to me that the ODBC setup on Windows might use different defaults than the actual driver defaults. However, that appears not to be the case. I commented out all but the "Big 5" (Driver, Servername, Database, UID, and PWD) and the test still passed.

2. `Debug=2` enables PostgreSQL ODBC logging. Like ODBC DM tracing, it slows things down and leaves "junk" in the target folder (C:\ on Windows, /tmp on Linux). Might as well leave it at the default (`Debug=0`) unless you really need it since it doesn't affect the outcome of the test anyway.

keitherskine:
Thanks for the info about the driver settings, Gord.  On the question of your earlier basic test, yes, it's curious how some things work and others don't.  Different drivers seem to encode and decode text in a multitude of ways, which makes things complex for pyodbc under the hood.  If Kleehammer got a few more grey hairs figuring it all out, I wouldn't be surprised.

Right now, I'm kinda keen to get to the point where all the tests pass (or at least the ones for MSSQL, PostgreSQL, and MySQL) on both Windows and at least some flavour of Linux.  I'll raise a ticket with PostgreSQL as you suggested and raise a small PR for that unit test change.  (Then in due course I'll raise another PR for Travis...)

keitherskine:
For what it's worth, I've raised an issue with PostgreSQL, number 16469:
https://www.postgresql.org/message-id/16469-11c82a64f17f51f4%40postgresql.org

I'm having a re-think about updating the unit test in question, though.  Maybe a narrower fix for Travis might be better.

Issue #617, opened 2019-09-02
Unable to insert UTF-8 character with fast_executemany option


maimat1:
I am trying to copy data from one SQL Server instance to another using pyODBC package and I encountered an error while handling UTF-8 characters on Windows.

I have narrowed it down to this snippet, which fails for me:
```python
import pyodbc
import sys
print('system:' + str(sys.getwindowsversion()))
print('python:' + sys.version)
print('pyodbc:' + pyodbc.version)
cnxn = pyodbc.connect(
    r'''Driver=ODBC Driver 17 for SQL Server;Server=localhost;'''
    r'''Database=DB_Pros;port=1433;Trusted_Connection=yes'''
    )
crsr = cnxn.cursor()
crsr.fast_executemany=True

crsr.execute("""DROP TABLE IF EXISTS dbo.testing_table""")
crsr.execute("""CREATE TABLE dbo.testing_table (col1 nvarchar(max) null)""")
crsr.executemany("""INSERT INTO dbo.testing_table (col1) VALUES (?)""", [[' 🎥 ',]])

return_value = crsr.execute("SELECT * FROM dbo.testing_table").fetchone()

print(return_value )

crsr.close()
cnxn.close()
```

I get the following error:
```
system:sys.getwindowsversion(major=10, minor=0, build=18362, platform=2, service_pack='')
python:3.7.1 (default, Oct 28 2018, 08:39:03) [MSC v.1912 64 bit (AMD64)]
pyodbc:4.0.27
Traceback (most recent call last):
  File "test.py", line 15, in <module>
    crsr.executemany("""INSERT INTO dbo.testing_table (col1) VALUES (?)""", [[' 🎥 ',]])
pyodbc.DataError: ('22026', '[22026] [Microsoft][ODBC Driver 17 for SQL Server]String data, length mismatch (0) (SQLParamData)')
```
Everything works fine when I set the `crsr.fast_executemany=False`.

This seems to be somewhat related to #246 , but this happens on Windows system.

v-chojas:
Could you post an ODBC trace?

maimat1:
Hi, thank you for looking into this.

Below attached is the ODBC trace, I have executed the snippet posted in the initial comment:
[SQL.LOG](https://github.com/mkleehammer/pyodbc/files/3579018/SQL.LOG)

v-makouz:
[...]

PR #652, opened 2019-11-27
Fix for #617 Incorrect size calculation for Unicode characters

v-makouz:
The bug manifests when using high Unicode characters that are more then 2 codepoints


tests3/sqlservertests.py
```diff
@@ -462,6 +462,13 @@ def test_fast_executemany_to_datetime2(self):
         self.cursor.executemany(sql, params)
         self.assertEqual(self.cursor.execute("SELECT CAST(dt2 AS VARCHAR) FROM ##issue540").fetchval(), '2019-03-12 10:00:00.12')
 
+    def test_high_unicode(self):
+        v = "🎥"
```
keitherskine: This is essentially cosmetic, but would it make it clear that this is a high unicode character if it was defined as `v = "\U0001F3A5"  # Movie Camera Emoji U+1F3A5` rather than just `v = "🎥"`?

mkleehammer: I agree.  Since I'd like to make a release immediately, I'll accept it as is and we can make a change afterwards.

Issue #175, opened 2017-01-06
Manylinux wheels


kevin-brown:
http://pythonwheels.com/

Right now, only Windows and Mac wheels are being distributed for [pyodbc on PyPI](https://pypi.python.org/pypi/pyodbc). For Linux, you need to compile billiard from source, which can add some time to getting everything up and running.

Support for Linux wheels was added in [PEP-513](https://www.python.org/dev/peps/pep-0513/) through [manylinux](https://github.com/pypa/manylinux). It attempts to make it easier to release wheels for the many Linux distributions out there, and the wheels can be generated automatically and in isolation through Docker containers.

I'm not sure what the current packaging process looks like, or if wheels can even pull in the required odbc libraries, so I can't speak for how feasible building manylinux wheels would be.

graingert:
@mkleehammer any update on this? would manylinux2010 help?

PR #668, opened 2020-01-03
Build Linux wheels on Travis with cibuildwheel

DavidMStraub:
I ran into #175 and thought I'd try fixing it by building Linux wheels directly on Travis CI using `cibuildwheel`. I tried this out in a fork and I believe it works.

You would have to add your `TWINE_USERNAME` and `TWINE_PASSWORD` as (encrypted) environment variables [in the repository settings](https://github.com/mkleehammer/pyodbc/settings/secrets).

I added the PyPI upload as `after_success`, so it will not affect the build status if there is a problem with the upload, unless there is a timeout (which happens when the environment vaiables are not set).

The only remaining problem I encountered is that PyPI reports
```
HTTPError: 400 Client Error: '4.0.29b9+commitf5ea676' is an invalid value for Version.
Error: Can't use PEP 440 local versions.
See https://packaging.python.org/specifications/core-metadata
for url: https://test.pypi.org/legacy/
```

But I believe this only happens when pushing a wheel for an untagged commit, which is anyway something one doesn't want.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,12 @@
 language: python
-sudo: false
+
+matrix:
+  include:
+  - sudo: required
+    services:
+      - docker
+    env: PIP=pip
+
 python:
   - "2.7"
   - "3.4"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,5 +1,12 @@
 language: python
-sudo: false
+
+matrix:
+  include:
+  - sudo: required
+    services:
+      - docker
+    env: PIP=pip
```
abitrolly: The `matrix` block is not needed. `$PIP` usage below should be replaced with `python -m pip install cibuildwheel twine` to prevent Travis failure.

abitrolly:
@DavidMStraub you may want to use https://config.travis-ci.com/explore to see how Travis interprets yaml.

PR #671, opened 2020-01-06
Very large integers and non-numeric floats as parameters

### LINKED ISSUE #670: Very large integers as parameters
### Environment

- Python: 3.7.5
- pyodbc: 4.0.28
- OS: Windows 10 Pro, 64-bit
- DB: SQL Server 2019 Express
- driver: ODBC Driver 17 for SQL Server

### Issue

Python allows integers of more-or-less any size.  When inserting a very large integer to a table as a parameter (a value larger than an 8 byte integer can contain), the value of -1 is inserted instead.  Oddly, an exception is also raised but not one of the standard PEP249 exceptions.  I would expect a PEP249 exception to be raised by pyodbc itself and nothing written to the database.  Here is the code that reproduces the issue:

```python
import pyodbc
cnxn = pyodbc.connect(dsn='<my_DSN>', autocommit=True)
crsr = cnxn.cursor()

crsr.execute("DROP TABLE IF EXISTS T1")
crsr.execute("CREATE TABLE T1 (col1 BIGINT NULL)")

try:
    sql = "INSERT INTO T1 (col1) VALUES (?)"
    parameters = (99999999999999999999999999999999999999999999999999,)
    crsr.execute(sql, parameters)
except Exception as err:
    print('exception:', err)

rows = crsr.execute("SELECT * FROM T1").fetchall()
print('table records:', rows)

crsr.close()
cnxn.close()
```

This code outputs:
```
exception: <method 'execute' of 'pyodbc.Cursor' objects> returned a result with an error set
table records: [(-1, )]
```

My guess is that [PyLong_AsLongLong](https://docs.python.org/3/c-api/long.html#c.PyLong_AsLongLong) is returning -1 (as expected) but then PyErr_Occurred() is not being checked somewhere.  [Here](https://github.com/mkleehammer/pyodbc/blob/master/src/params.cpp#L1469) perhaps?

keitherskine:
This PR attempts to fix #670 .  Currently, conversions from PyLongObject and PyFloatObject type python parameters to C-style values are not checked for success.  This PR addresses that.

When converting integers, an Overflow exception is generated by [PyLong_AsLongLong()](https://docs.python.org/3/c-api/long.html#c.PyLong_AsLongLong) if the integer value is larger than an 8-byte integer can contain.  For floats, no exception is generated by [PyFloat_AsDouble()](https://docs.python.org/3/c-api/float.html#c.PyFloat_AsDouble) for non-numeric float values, but a ProgrammingError is raised by the driver.

It may be better in future to raise DataError exceptions in both cases, but that would require more extensive code changes.

### DIFF
--- a/src/params.cpp
+++ b/src/params.cpp
@@ -858,21 +858,23 @@ static bool GetTimeInfo(Cursor* cur, Py_ssize_t index, PyObject* param, ParamInf
 }
 
 
-inline bool NeedsBigInt(PyObject* p)
+inline bool NeedsBigInt(long long ll)
 {
     // NOTE: Smallest 32-bit int should be -214748368 but the MS compiler v.1900 AMD64
     // says that (10 < -2147483648).  Perhaps I miscalculated the minimum?
-    long long ll = PyLong_AsLongLong(p);
     return ll < -2147483647 || ll > 2147483647;
 }
 
 #if PY_MAJOR_VERSION < 3
 static bool GetIntInfo(Cursor* cur, Py_ssize_t index, PyObject* param, ParamInfo& info, bool isTVP)
 {
-    if (isTVP || NeedsBigInt(param))
-    {
-        info.Data.i64 = (INT64)PyLong_AsLongLong(param);
+    long long value = PyLong_AsLongLong(param);
+    if (PyErr_Occurred())
+        return false;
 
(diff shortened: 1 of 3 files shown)


mkleehammer:
Great catch - thanks.  I like the tests.

PR #689, opened 2020-01-25
Add a note about C++ compilers to README.md

keitherskine:
This updates the README text to include the need for a suitable C++ compiler when installing pyodbc, and a reference to the appropriate wiki page as well.  Hopefully this will help people when installing pyodbc for the first time.


mkleehammer:
Great idea - I still field direct email questions on this all the time.

PR #690, opened 2020-01-25
Use README.md for the long description in setup.py

keitherskine:
On the PyPi website, the project description is derived from the "long description" defined in `setup.py`.  In pyodbc's case, the "long description" is currently quite [short](https://pypi.org/project/pyodbc/#description).  This PR changes `setup.py` to use the contents of the README.md file for the "long description" instead.  This will help to provide more information to users, particuarly those wishing to install pyodbc for the first time.

Also, this PR makes another small fix to `setup.py`.  One of the regex strings is not defined as a "raw" string, yet contains the "\d" regex [character combination](https://docs.python.org/3/library/re.html#regular-expression-syntax).  This currently [raises](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals) a DeprecationWarning in later versions of Python.  The regex string just needs to be made "raw" which I have done.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,6 +67,9 @@ def main():
 
     version_str, version = get_version()
 
+    with open(join(dirname(abspath(__file__)), 'README.md')) as f:
+        long_description = f.read()
+
     settings = get_compiler_settings(version_str)
 
     files = [ relpath(join('src', f)) for f in os.listdir('src') if f.endswith('.cpp') ]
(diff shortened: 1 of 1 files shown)


mkleehammer:
I had not thought of that.   Interesting idea.  I've noticed some projects with what looked like their readme and assumed that was an option somewhere in PyPI I couldn't find.

mkleehammer:
I just saw the results after publishing 4.0.30: https://pypi.org/project/pyodbc.  That was a very nice improvement.  Thanks.

keitherskine:
No probs, glad it helped.

PR #692, opened 2020-01-25
Compile for venv

khalidmammadov:
Installing the package into virtual environment fails due to missing headers in POSIX systems where compilation is necessary. This will include path where headers are stored as installed by ibpython3.X-dev module

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -197,7 +197,7 @@ def get_compiler_settings(version_str):
         dirs = [item.path for item in os.scandir('/usr/include')
                 if item.name.startswith('python') and item.is_dir()]
         settings['include_dirs'].extend(dirs)
-        
+
         cflags = os.popen('odbc_config --cflags 2>/dev/null').read().strip()
         if cflags:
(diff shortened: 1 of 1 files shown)


mkleehammer:
This seems like it is going to add the all Python versions, so the compiler would only pick up the first header file it finds.  That might cause build problems.  Is there a way to find the headers for the matching version?

Also, I really thought distutils did that automatically.  I'll have to spin up a new VM and try to reproduce this problem.  Does it happen every time you try to install into a venv on Linux?

khalidmammadov:
That makes sense to handle versions. 
Actually, I have just tried to install it on ubuntu docker image without VENV and it fails.
Also, trying official python docker image and fails again. Will try on Alpine as well and let you know.

khalidmammadov:
Checked on Alpine and the same.
The only control I can suggest is to check python version and append major.manor to "python" folder name string i.g. `if item.name.startswith('python3.8')`

khalidmammadov:
Not needed

PR #696, opened 2020-02-02
Return an exit code from each test script

keitherskine:
When running the test scripts from the command line, it's useful to be able to tell programmatically if they succeeded or not.  To make this possible, this PR adds a `sys.exit()` call at the end of the `__main__` section of all the database test scripts.  The test scripts now return 0 if all the tests pass, 1 otherwise.

I've used the TestResult [wasSuccessful()](https://docs.python.org/3/library/unittest.html#unittest.TestResult.wasSuccessful) function to determine if the tests passed or not.  Happy to use an alternate function if preferred.

### DIFF
--- a/tests2/accesstests.py
+++ b/tests2/accesstests.py
@@ -659,10 +659,12 @@ def main():
     testRunner = unittest.TextTestRunner(verbosity=options.verbose)
     result = testRunner.run(suite)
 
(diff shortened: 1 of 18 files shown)


mkleehammer:
Great idea.  Thanks.

PR #697, opened 2020-02-02
Run unit tests with AppVeyor CI

keitherskine:
Note, this PR has a dependency on PR #705 .

This PR is an attempt to introduce automated building and testing of pyodbc commits (on Windows), across a range of Python versions and databases.  With this update, AppVeyor CI will run the existing unit tests on Windows AppVeyor servers for the following databases and ODBC drivers:

- MS SQL Server 2017
  - {SQL Server Native Client 10.0} (if available on the server)
  - {SQL Server Native Client 11.0} (if available on the server)
  - {ODBC Driver 11 for SQL Server}
  - {ODBC Driver 13 for SQL Server}
  - {ODBC Driver 17 for SQL Server} 
- PostgreSQL 11
  - {PostgreSQL Unicode} (32-bit Python)
  - {PostgreSQL Unicode(x64)} (64-bit Python)
- MySQL 5.7
  - {MySQL ODBC 5.3 ANSI Driver} (Python 2.7)
  - {MySQL ODBC 8.0 ANSI Driver} (Python 3)

The existing unit tests will be run against Python versions 2.7, 3.5, 3.6, 3.7, and 3.8.

Most of the functionality is accomplished through DOS/Powershell batch scripts.  The main one is the DOS script that runs the tests (`appveyor\test_script.cmd`).  It's MS-DOS so is kinda long-winded and clunky but it does the job.  Building the pyodbc module, especially for Python 2.7, is facilitated using a script (`appveyor\compile.cmd`) created by Olivier Grisel.  I'm no expert on compiling C++ in a Windows environment so I'll be happy and grateful for any feedback on this approach.

There are various environmental variables that can be set to control which databases are tested and whether "wheel" distributable files are generated.  See the `appveyor.yml` file for details.  Wheel files are made available as AppVeyor artifacts, downloadable through the AppVeyor UI.  You will need to [sign up](https://ci.appveyor.com/signup) for an AppVeyor account for this, and probably fork the pyodbc repo too.

Some ODBC drivers are pre-installed in the AppVeyor VM images.  Some need to be installed at the start of each job.  These drivers are downloaded from their source if necessary, but to make the CI process quicker and more reliable, the AppVeyor [cache](https://www.appveyor.com/docs/build-cache/) mechanism is used to store the msi files between builds.

[...]

PR #698, opened 2020-02-04
SyntaxError fix and other changes to some test scripts

keitherskine:
This PR is related to PR #697 (the AppVeyor change).  Contrary to what I indicated earlier, it's best if this PR could be merged first, because it fixes and tidies up some things in the unit tests scripts used by that PR.

Various test scripts have been updated with the following changes:

1. Fix the SyntaxError issue for Python 3 test scripts that are coded in "latin-1".  There is a bug in Python 3.4, 3.5, and 3.6 which generates a SyntaxError when reading the coding.  It seems to be related to [this](https://bugs.python.org/issue20731) Python issue.
2. Call `print_library_info()` only when the "verbose" flag is set.  This has already been done on some test scripts, and it helps to keep the test output clean.
3. Convert the "usage" string defined at the start of some scripts to the module's [docstring](https://www.askpython.com/python/python-docstring).  Then the module docstring can be referenced in the rest of the module as simply `__doc__`.  A cosmetic change, it's true, but I believe it makes the code more Pythonesque.
4. I also took the liberty of switching from the `OptionParser` class to the `ArgumentParser` class for extracting the command line arguments.  The module `optparse` has been [deprecated](https://docs.python.org/3.8/library/optparse.html) since Python 3.2 so it would seem worthwhile to upgrade the code.
5. Finally, some minor formatting of the usage instructions.

Happy to back out any of these changes if necessary.

### DIFF
--- a/tests2/mysqltests.py
+++ b/tests2/mysqltests.py
@@ -736,9 +736,10 @@ def main():
             parser.print_help()
             raise SystemExit()
 
-    cnxn = pyodbc.connect(connection_string)
-    print_library_info(cnxn)
-    cnxn.close()
+    if args.verbose:
+        cnxn = pyodbc.connect(connection_string)
+        print_library_info(cnxn)
+        cnxn.close()
 
     suite = load_tests(MySqlTestCase, args.test, connection_string)
 
--- a/tests2/mysqltests.py
+++ b/tests2/mysqltests.py
@@ -1,21 +1,21 @@
 #!/usr/bin/python
 # -*- coding: latin-1 -*-
 
-usage = """\
-usage: %prog [options] connection_string
-
+"""
 Unit tests for MySQL.  To use, pass a connection string as the parameter.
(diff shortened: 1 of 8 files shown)


keitherskine:
Just FYI, I notice the existing AppVeyor build is failing for this PR with the error:
```
fatal error C1007: unrecognized flag '-FH4-' in 'p2'
error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\BIN\\cl.exe' failed with exit status 1
```
This is probably because the default Windows Image for AppVeyor is 2015 (I believe).  For Python 3.6, Windows Image 2019 is now needed after the recent `setup.py` [change](https://github.com/mkleehammer/pyodbc/commit/b4ea03220dd8243e452c91689bef34823b2f7d8f).

keitherskine:
I'm closing this PR in favor of PRs #705 and #706 .  Better to split it up.

PR #705, opened 2020-02-10
Fix SyntaxError error for non-utf8 coded files

keitherskine:
There is a bug in Python 3.4, 3.5, and 3.6 which generates a SyntaxError when the coding declaration in the header (e.g. `# -*- coding: latin-1 -*-`) is read by the Python interpreter.  The issue occurs when the files have unix (line-feed) line endings and the coding is not "utf-8".  It seems to be related to Python issues [20731](https://bugs.python.org/issue20731), [20844](https://bugs.python.org/issue20844), [27797](https://bugs.python.org/issue27797), and [32809](https://bugs.python.org/issue32809).  This PR provides a workaround to that issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,15 +1,55 @@
+# This file is based largely on the GitHub .gitignore template for Python:
+# https://github.com/github/gitignore/blob/master/Python.gitignore
+
+# Python
+__pycache__/
+*.py[cod]
 setup.cfg
+
+# C extensions
+*.so
+*.dll
+
+# Distribution / packaging
+build/
+dist/
+lib/
+lib64/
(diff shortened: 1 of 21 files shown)


mkleehammer:
Would it be better to convert these to utf-8?  They are all Python 3 test files, so the source encoding won't matter as much - they will be converted properly.

I'd prefer to do this, but your other PR for automated testing relies on this so I don't want to make changes until we discuss it.  Thanks.

keitherskine:
Yes, converting the files to utf-8 would be better and I have done so.

I will update PR 697 to align with this change.

(By the way, @mkleehammer, great to have you back!)

PR #719, opened 2020-02-29
Fix DSN truncation in dataSources()

### LINKED ISSUE #718: dataSources() DSN truncation
### Environment

- Python: 3.7.6 (64-bit)
- pyodbc: 4.0.30
- OS: Windows 10
- DB: MS SQL Server 2012
- Driver: ODBC Driver 17 for SQL Server

### Issue

Observed behavior:
- `dataSources()` truncates DSN to 31 characters if DSN is 32 characters long

Expected behavior:
- `dataSources()` does not truncate DSN if it's 32 characters long 

Suspected cause:
`szDSN` array in `mod_datasources()` doesn't take into account extra space needed for null-terminator, therefore DSN names that utilize maximum available space of 32 characters (defined by `SQL_MAX_DSN_LENGTH`) get truncated.

Proposed solution:
Declare `szDSN` array using `SQL_MAX_DSN_LENGTH + 1` to accommodate extra space needed for null-terminator.

qpenko:
Updated buffer array declaration to accommodate extra space for null terminator and therefore allow DSN to utilize all available space of 32 characters without truncation.

Fixes #718

### DIFF
--- a/src/pyodbcmodule.cpp
+++ b/src/pyodbcmodule.cpp
@@ -617,7 +617,7 @@ static PyObject* mod_datasources(PyObject* self)
     if (!result)
         return 0;
 
-    SQLCHAR szDSN[SQL_MAX_DSN_LENGTH];
+    SQLCHAR szDSN[SQL_MAX_DSN_LENGTH + 1];
     SWORD cbDSN;
     SQLCHAR szDesc[200];
     SWORD cbDesc;
--- a/src/pyodbcmodule.cpp
+++ b/src/pyodbcmodule.cpp
@@ -617,9 +617,9 @@ static PyObject* mod_datasources(PyObject* self)
     if (!result)
         return 0;
 
-    SQLCHAR szDSN[SQL_MAX_DSN_LENGTH + 1];
+    SQLCHAR szDSN[500]; // Using a buffer larger than SQL_MAX_DSN_LENGTH + 1 for systems that ignore it
     SWORD cbDSN;
-    SQLCHAR szDesc[200];
+    SQLCHAR szDesc[500];
     SWORD cbDesc;
 
     SQLUSMALLINT nDirection = SQL_FETCH_FIRST;


keitherskine:
Good catch @qpenko .  This will fix things for Windows DSNs, where the DSN name apparently has to be in ASCII (i.e. one-byte) characters.

Just a thought though.  For Unix systems, the DSN name will probably be utf-8, where each character can theoretically be up to 4 bytes.  Perhaps the buffer should have 4 bytes per character, plus the null-byte character, just in case.  Something like `SQLCHAR szDSN[(SQL_MAX_DSN_LENGTH * 4) + 1];` ?

Or am I mistaken that SQL_MAX_DSN_LENGTH refers to code points, as opposed to bytes?

See [docs](https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqldatasources-function?view=#arguments) ("BufferLength1" entry) and [sqlext.h](https://github.com/microsoft/ODBC-Specification/blob/master/Windows/inc/sqlext.h#L45) for reference.

qpenko:
Fair point. I used to think SQL_MAX_DSN_LENGTH was referring to bytes, but now that you've mentioned other OSs I'm not really sure. What do you think we better do? Would it make sense to perform a simple check on a *nix system?

keitherskine:
OK, so I did an experiment on a Linux 16.04 machine.  I changed the relevant line in pyodbc to:
`SQLCHAR szDSN[1000];`
to simply create a large buffer for testing.  I updated my `odbc.ini` file to include a long unicode name, thus:
```
[₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼ABC]
Description = MS SQL Server
Driver      = ODBC Driver 13 for SQL Server
Server      = localhost
```
₼, I'm sure you'll recognize is the currency symbol for Azerbaijan, unicode symbol U+20BC.  I then ran the following:

```
$ python3 -c "import pyodbc; print(pyodbc.dataSources())"
{'₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼₼ABC': 'ODBC Driver 13 for SQL Server'}
$
$ python3 -c "import pyodbc; d = pyodbc.dataSources(); n = list(d.keys())[0]; print(n.encode('utf-8'), len(n), len(n.encode('utf-8')))"
b'\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbc\xe2\x82\xbcABC' 63 183
```
As you can see, the data source name returned is much longer than the 32 characters defined in SQL_MAX_DSN_LENGTH.  In this case it's 63 characters long and 183 bytes when encoded in utf-8.

My conclusion is that the szDSN buffer should not be based on SQL_MAX_DSN_LENGTH at all, despite what the docs say.  I would just hard-code it to something large, like 250 or 500 or 1000 bytes.  Perhaps put a comment in there too to explain, e.g. "Some drivers ignore SQL_MAX_DSN_LENGTH so just using a large buffer instead".

keitherskine:
Whilst we're here, perhaps there's a case for aligning the buffers for the driver descriptions in lines 569 and 622 of `pyodbcmodule.cpp`.  Currently, one is 200 bytes and the other is 500 bytes.  Perhaps they should both be made 500 bytes long?

keitherskine:
That looks good to me.

qpenko:
You were right about the buffer size! Thank you for running the experiment on Linux. I've updated all buffer sizes to 500 bytes. Now that they are all aligned, maybe we could tie them to a constant for easier reference and update? We could either create a new one, or derive from existing (e.g. SQL_MAX_DSN_LENGTH * SQL_MAX_DSN_LENGTH for 1024 bytes, though it may not be a good idea to abuse this value in such way).

Issue #495, opened 2018-12-14
Retrieving informational messages returned by server


gordthompson:
(Inspired by [this Stack Overflow question](https://stackoverflow.com/q/53778988/2144390).)

Is there a way that we can retrieve informational messages sent by the server, for example, those produced by a `PRINT` statement in a T-SQL stored procedure?

For a test SP

```sql
USE [myDb]
GO

/****** Object:  StoredProcedure [dbo].[PrintTest]    Script Date: 2018-12-14 10:59:12 ******/
SET ANSI_NULLS ON
GO

SET QUOTED_IDENTIFIER ON
GO

-- =============================================
-- Author:		gord
-- Create date: 
-- Description:	
-- =============================================
ALTER PROCEDURE [dbo].[PrintTest] 
AS
BEGIN
    SET NOCOUNT ON;
    PRINT 'no es bueno';
END

GO
```

the message does appear in the ODBC trace

```
main            1c0c-18b0	ENTER SQLExecDirectW 
        HSTMT               0x000000000040DFA0
        WCHAR *             0x0000000002F52830 [      18] "EXEC dbo.PrintTest"
        SDWORD                    18

main            1c0c-18b0	EXIT  SQLExecDirectW  with return code 1 (SQL_SUCCESS_WITH_INFO)
        HSTMT               0x000000000040DFA0
        WCHAR *             0x0000000002F52830 [      18] "EXEC dbo.PrintTest"
        SDWORD                    18

        DIAG [01000] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]no es bueno (0) 
```

but is there a way we can retrieve it in our Python script?

mkleehammer:
I don't see why not - isn't that just another diagnostics record just like we get for exceptions?  I could see having a method to return an array of diagnostic messages.

What format would be best for returning them.  In the spec, they have SQLSTATE codes, text, and native error codes.

If I was using the API, I'd just want the text if I knew it was a print statement, but nothing here really tells us that.  SQLSTATE 01000 is a warning with no further classification IIRC.  I suppose either a new object or a tuple.  An object would allow us to extend it in the future.  (It needs to be something available in Python 2.7 until we finally stop supporting that.  It's EOL is coming up soon.)

gordthompson:
My suggestion would be to return the diagnostic records in the most complete and granular form, perhaps a list of tuples something like

```
[(Msg, SqlState, NativeError), ...]
```

[...]

PR #765, opened 2020-05-22
New cursor "messages" attribute to retrieve PRINT statements

keitherskine:
Here is my attempt to implement a mechanism for retrieving errors, warnings, and PRINT statements from the database in the manner [described](https://www.python.org/dev/peps/pep-0249/#cursor-messages) in PEP-249, i.e. through the use of a "messages" attribute on cursor objects.  This functionality is being tracked in issue #495 .

Currently, it is not possible for pyodbc clients to retrieve any errors, warnings, or other messages that have been generated by the database.  For example, `PRINT 'Start of Proc'` outputs a message, but pyodbc currently ignores it.  These messages are made available by the driver manager as "diagnostic records" through the [SQLGetDiagRec](https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqlgetdiagrec-function) function.

PEP-249 [suggests](https://www.python.org/dev/peps/pep-0249/#cursor-messages) these messages should be made available as a list of tuples, where each tuple is made up of an "exception class" and an "exception message".  I'm not sure what that "exception class" is supposed to be, so for the moment I am simply including the SQLSTATE and the error number in the first element of the tuple, formatted in alignment with the current exception message format, and including the error message in the second element, e.g.:
```python
crsr.execute("PRINT 'hello world'")
print(crsr.messages)
```
would output:
```
[('[01000] (0)', '[Microsoft][ODBC Driver 17 for SQL Server][SQL Server]hello world')]
```

Each time `nextset()` is called, the messages are deleted and replaced with new messages from the result set (if any).  It should be noted, each result set can include both row data and messages.  It's also possible to have a result set that includes messages but does not have any row data.  (This isn't a change in behavior, it's always been possible to have a results set with no actual data records.)  In that scenario, clients can read any messages with `crsr.messages` but if they try to call `fetchall()`, an exception is generated.  This behavior is [per PEP-249](https://www.python.org/dev/peps/pep-0249/#fetchall).  Personally, I think it might have been easier if `fetchall()` just returned `None` if there weren't any records rather than raise an exception, but that's how PEP-249 is written.  On a brand new cursor object, the cursor message attribute starts out as `None`.  After calling `execute()` or `nextset()`, if no messages are available, an empty list is returned.

As part of this work, I exposed the method `CopySqlState()` by moving it from `errors.cpp` to `errors.h` so that it is available from `cursor.cpp`.

Memory management was checked using the [psutil](https://github.com/giampaolo/psutil) library.

A quick word on unit tests for the various databases.  SQL Server has a built-in [PRINT](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/print-transact-sql) statement, so that's relatively easy to test.  PostgreSQL doesn't support PRINT but has [RAISE INFO](https://www.postgresql.org/docs/11/plpgsql-errors-and-messages.html) instead.  For multiple RAISE INFO statements though, it appears only the last one is returned to the client.  As far as I can tell, MySQL doesn't have any way of returning commentary to the client.

Any and all thoughts welcome.

### DIFF
--- a/src/cursor.cpp
+++ b/src/cursor.cpp
@@ -300,6 +300,7 @@ enum free_results_flags
     KEEP_STATEMENT = 0x02,
     FREE_PREPARED  = 0x04,
     KEEP_PREPARED  = 0x08,
+    KEEP_MESSAGES  = 0x10,
 
     STATEMENT_MASK = 0x03,
     PREPARED_MASK  = 0x0C
@@ -365,6 +366,12 @@ static bool free_results(Cursor* self, int flags)
         self->map_name_to_index = 0;
     }
 
+    if ((flags & KEEP_MESSAGES) == 0)
+    {
+        Py_XDECREF(self->messages);
+        self->messages = PyList_New(0);
+    }
+
     self->rowcount = -1;
 
     return true;
@@ -401,11 +408,13 @@ static void closeimpl(Cursor* cur)
     Py_XDECREF(cur->description);
     Py_XDECREF(cur->map_name_to_index);
     Py_XDECREF(cur->cnxn);
+    Py_XDECREF(cur->messages);
 
     cur->pPreparedSQL = 0;
     cur->description = 0;
     cur->map_name_to_index = 0;
     cur->cnxn = 0;
+    cur->messages = 0;
 }
 
 static char close_doc[] =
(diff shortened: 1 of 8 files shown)


mkleehammer:
Wow.  That code is a pleasure to read!  Well done.

falonso-caa:
Wow, this is very exciting! Looking forward to the release.

Fredchen717:
I just checked it in my deployment pipeline. Everything works. Can't wait for the release.

ronak2303:
when is this getting released, we are looking forward of this cur.messages object.

v-chojas:
Only @mkleehammer can answer that.

MylesFowler:
Bookmarked and watching. This would be awesome.

bfmcneill:
any suggestion on how to get this to bubble up to sqlalchemy when using pyodbc?

v-chojas:
You'll have to ask sqlalchemy to add code to read this attribute.

iwccsbfb:
This can only retrieve one print message, when I executed two print messages with sql server, it can only show the first one.

keitherskine:
@iwccsbfb Each `.execute` is a separate context, but if the PRINT statements are in the same SQL, they should be visible using `.nextset()`:

```python
import pyodbc
cnxn = pyodbc.connect(my_cnxn_str, autocommit=True)
crsr = cnxn.cursor()

crsr.execute("PRINT 'AAA'")
crsr.execute("PRINT 'BBB'")
print(crsr.messages)   # shows only BBB
print(crsr.nextset())  # returns False

crsr.execute("PRINT 'CCC'; PRINT 'DDD'")
print(crsr.messages)  # shows only CCC
print(crsr.nextset()) # returns True
print(crsr.messages)  # shows only DDD
print(crsr.nextset()) # returns False

crsr.close()
cnxn.close()
```
output:
```
[('[01000] (0)', '[Microsoft][ODBC Driver 17 for SQL Server][SQL Server]BBB')]
False
[('[01000] (0)', '[Microsoft][ODBC Driver 17 for SQL Server][SQL Server]CCC')]
True
[('[01000] (0)', '[Microsoft][ODBC Driver 17 for SQL Server][SQL Server]DDD')]
False
```

iwccsbfb:
@keitherskine Thank you for the reply. This would work for me for multiple print statements, however, for non-print output, do you know how to capture it? For example, the update statement below would show 
```(1 row affected)
Completion time: 2021-12-06T15:31:55.0032644-05:00```  
If I run in sql server or sqlcmd. However, I am not able to capture it with pyodbc.


```
code1 = '''
print ('output')
print ('output1')

UPDATE dbo."tbXXXX"
SET Name = 'XXX'
WHERE rowID=4

print ('output2')
print ('output3')
'''

if __name__ == '__main__':
    conn = pyodbc.connect(config)
    print (pyodbc.version)
    cur = conn.cursor()
    cur.execute(code1)
    print(cur.messages)
    while cur.nextset():
        print( cur.messages )
```

```

[('[01000] (0)', '[Microsoft][ODBC Driver 17 for SQL Server][SQL Server]output')]
[('[01000] (0)', '[Microsoft][ODBC Driver 17 for SQL Server][SQL Server]output1')]
[]
[('[01000] (0)', '[Microsoft][ODBC Driver 17 for SQL Server][SQL Server]output2')]
[('[01000] (0)', '[Microsoft][ODBC Driver 17 for SQL Server][SQL Server]output3')]
```

keitherskine:
@iwccsbfb I could be wrong, but as I understand it, output like `(1 row affected) Completion time: 2021-12-06T15:31:55.0032644-05:00` is not directly part of the ODBC API.  It is output by the client application (SSMS, sqlcmd, etc.).  The number of rows comes from `.rowcount` and the completion time is, well, the time the SQL completes.  Hence, I believe you should be able to re-create that output if you really need it.

PR #766, opened 2020-05-23
handle NullParam like None in ExecuteMulti

may-day:
if fast_executemany is True NullParam was not recognized


mkleehammer:
Great catch.

PR #767, opened 2020-05-24
SQL Server 2019 change of data truncation error

keitherskine:
In SQL Server 2019, the error message generated for data truncation errors has [changed](https://support.microsoft.com/en-us/help/4468101/optional-replacement-for-string-or-binary-data-would-be-truncated) (for the better!).  In SS2019 this raises a `pyodbc.ProgrammingError` exception rather than a `pyodbc.DataError` exception, which breaks some unit tests when run on SS2019.  This PR changes the relevant unit tests to check for `pyodbc.DataError` or `pyodbc.ProgrammingError`, so that the tests pass for both SQL Server 2019 and older SQL Server instances.


keitherskine:
This fix is included in #773 .

PR #776, opened 2020-06-05
Bind TVP columns even when the table is empty

### LINKED ISSUE #772: Invocation of stored procedure with empty TVP dumps core
### Environment

- Python: 3.8.2
- pyodbc: 4.0.30
- OS: Linux 5.4.0-31-generic (also reproduced on Windows Server 2016)
- DB: SQL Server 2019 (RTM-CU4) (KB4548597) - 15.0.4033.1 (X64)
- driver: Microsoft ODBC Driver 17 for SQL Server

### Issue

- Expected: successful execution or a raised exception
- Observed: segmentation fault (core dumped)

bkline:
If the columns of a table-valued parameter are not bound, the
Python interpreter will crash, even if the value passed for
the invocation of the stored procedure has no rows. According
to Microsoft's documentation:

> After binding the table-valued parameter, the application
> must then bind each table-valued parameter column.
> 

Closes #772

### DIFF
--- a/src/cursor.h
+++ b/src/cursor.h
@@ -68,6 +68,9 @@ struct ParamInfo
     struct ParamInfo *nested;
     SQLLEN curTvpRow;
 
+    // For TVPs, the name of the table type.
+    SQLWCHAR* type_name;
+
     // Optional data.  If used, ParameterValuePtr will point into this.
     union
     {
--- a/src/params.cpp
+++ b/src/params.cpp
@@ -606,6 +606,8 @@ static void FreeInfos(ParamInfo* a, Py_ssize_t count)
             pyodbc_free(a[i].ParameterValuePtr);
         if (a[i].ParameterType == SQL_SS_TABLE && a[i].nested)
             FreeInfos(a[i].nested, a[i].maxlength);
+        if (a[i].type_name)
+            pyodbc_free(a[i].type_name);
         Py_XDECREF(a[i].pObject);
     }
     pyodbc_free(a);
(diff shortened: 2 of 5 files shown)


bkline:
Any thoughts about the pipeline check failures? Yes there are warnings (looks like they've been there for a while), but I'm able to build successfully both on Linux and on Windows.

keitherskine:
The Travis/AppVeyor build pipelines don't do much at the moment, and are kinda broken anyway, but hopefully this should be remedied in the near future.

bkline:
Not directly related to this PR, but I noticed when I was running the unit tests in `sqlservertests.py` that all of the tests (including my modification of `test_tvp()`) pass on Windows, but `test_too_large()` fails on Linux, because on that platform the attempt to insert too much data into a column raises a `pyodbc.ProgrammingError` instead of the expected `pyodbc.DataError`.

keitherskine:
Yes, hence PR 767.

v-chojas:
You can just unbind the TVP parameter if there are no rows. The default is an empty TVP.

bkline:
@v-chojas  Please show me the patch for your approach. I don't see a call which would unbind a parameter in the ODBC API arsenal, but I have tried your suggestion of avoiding the binding altogether, by adding this at the top of `BindParemeter()` (right after the `UpdateParamInfo()` block):

```c++
    // Avoid binding an empty TVP.
    if (sqltype == SQL_SS_TABLE)
    {
        Py_ssize_t i = PySequence_Size(info.pObject);
        if (i < 0)
        {
            RaiseErrorV(0, ProgrammingError, "A TVP must be a Sequence.");
        }
        else if (!i)
            return true;
    }
```

That results in the following error when an empty TVP is passed to the stored procedure:

> COUNT field incorrect or syntax error (0) (SQLExecDirectW)')

bkline:
@v-chojas We are trying very hard to keep an open mind for your proposed solution, but it isn't going to be easy if you don't provide details about how that solution will work. In particular, your suggestion that we
> ... just unbind the TVP parameter if there are no rows

isn't supported by this language in the [Microsoft ODBC documentation](https://docs.microsoft.com/en-us/sql/odbc/reference/develop-app/binding-parameters-odbc?view=sql-server-ver15):
> A variable remains bound to a parameter until a different variable is bound to the parameter, until all parameters are unbound by calling SQLFreeStmt with the SQL_RESET_PARAMS option, or until the statement is released.

We haven't been able to find anything (beyond your suggestion) which contradicts that passage, or offers a function for unbinding a single bound parameter.

Furthermore, that same page in the documentation begins ...
> Each parameter in an SQL statement must be associated, or bound, to a variable in the application before the statement is executed.

The best way for you to demonstrate that the documentation is wrong and that your solution will work is to provide the solution so we can test it. Trust me, we'll be delighted if it works.

If you are unable or unwilling to do that, please test this PR, and merge it if it works correctly. Notice that the _Allow edits by maintainers_ flag is turned on, so you can modify the PR if you believe further changes are necessary.

Cheers and thanks,
Bob 😅

v-chojas:
I am only providing some general hints, because I know there is a simpler solution that does not involve an extra round-trip to the server, especially when there is no need in this scenario to even send any data in the first place, but since you insist...

Add `i >= 0` to the loop condition and `info.StrLen_or_Ind = SQL_DEFAULT_PARAM;` next to the `info.nested = 0;` (no column case).

I have no write access to this repository.

bkline:
Thanks, @v-chojas, that works. I have submitted a [replacement PR](https://github.com/mkleehammer/pyodbc/pull/779), and will close this one. I will file a bug report to get the [documentation](https://docs.microsoft.com/en-us/sql/relational-databases/native-client-odbc-table-valued-parameters/binding-and-data-transfer-of-table-valued-parameters-and-column-values?view=sql-server-ver15) fixed, where it says,

> After binding the table-valued parameter, the application must then bind each table-valued parameter column.

I appreciate your assistance.

bkline:
@v-chojas - Just out of curiosity, how did you know that the [Microsoft ODBC documentation](https://docs.microsoft.com/en-us/sql/relational-databases/native-client-odbc-table-valued-parameters/binding-and-data-transfer-of-table-valued-parameters-and-column-values?view=sql-server-ver15) was wrong?

> Table-valued parameters as a whole cannot be set to a default value by using SQL_DEFAULT_PARAM in StrLen_or_IndPtr ....

v-chojas:
I never said the documentation was wrong, and it is not. That particular sentence is referring to default values, not the absence of rows.

Look further down on that page and you will see:
"Sets StrLen_or_IndPtr or SQL_DESC_OCTET_LENGTH_PTR to SQL_DEFAULT_PARAM for a table-valued parameter _with no rows_."

PR #792, opened 2020-06-29
Don't use PyUnicode_Encode API

methane:
PyUnicode_EncodeUTF8 is deprecated, and PyUnicode_EncodeUTF8(PyUnicode_AS_UNICODE(p), ...) is inefficient.

### DIFF
--- a/src/params.cpp
+++ b/src/params.cpp
@@ -1921,9 +1921,10 @@ bool ExecuteMulti(Cursor* cur, PyObject* pSql, PyObject* paramArrayObj)
                 if (PyUnicode_Check(objCell))
                 {
                     const TextEnc& enc = cur->cnxn->sqlwchar_enc;
-                    int cb = PyUnicode_GET_DATA_SIZE(objCell) / 2;
-
                     PyObject* bytes = NULL;
+
+#if PY_MAJOR_VERSION < 3
+                    int cb = PyUnicode_GET_DATA_SIZE(objCell) / 2;
                     const Py_UNICODE* source = PyUnicode_AS_UNICODE(objCell);
 
                     switch (enc.optenc)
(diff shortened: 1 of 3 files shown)


src/params.cpp
```diff
@@ -1941,11 +1942,28 @@ bool ExecuteMulti(Cursor* cur, PyObject* pSql, PyObject* paramArrayObj)
...
+                    switch (enc.optenc)
+                    {
+                    case OPTENC_UTF8:
+                        bytes = PyUnicode_AsUTF8String(objCell);
+                        break;
+                    case OPTENC_UTF16:
+                        bytes = PyUnicode_AsUTF16String(objCell);
+                        break;
+                    case OPTENC_UTF16LE:
+                        bytes = PyUnicode_AsEncodedString(objCell, "utf_16_le", NULL);
```
keitherskine, line 1955: Minor, but it might be nice to specify the errors parameter as "strict" rather than leaving it as NULL.  Ditto for OPTENC_UTF16BE.

methane: In Python/C API, `errors=NULL` means "strict" consistently, like `errors=None` in Python.
And it is little faster/efficient because Python can skip `strcmp()`, creating temporary Unicode object, or looking up error handler from registry.

Do you prefer not using the default anyway?

fila43:
The build with python3.11 fails with: 
`src/cnxninfo.cpp:45:18: error: â€˜PyUnicode_EncodeUTF8â€™ was not declared in this scope; did you mean â€˜PyUnicode_DecodeUTF8â€™?
   45 |     Object bytes(PyUnicode_EncodeUTF8(PyUnicode_AS_UNICODE(p), PyUnicode_GET_SIZE(p), 0));
      |                  ^~~~~~~~~~~~~~~~~~~~
      |                  PyUnicode_DecodeUTF8`

The change proposed in this PR fixes this issue.

Sitaramjaat:
133

Sitaramjaat:
331517

Sitaramjaat:
Mast 12

Issue #802, opened 2020-08-06
Increasing memory consumption with fast_executemany and varchar(max)


gordthompson:
### Environment

To diagnose, we usually need to know the following, including version numbers.  On Windows, be
sure to specify 32-bit Python or 64-bit:

- Python: 3.7 x64
- pyodbc: 4.0.30
- OS: Windows 8.1
- DB: SQL Server
- driver: ODBC Driver 17 for SQL Server

### Issue
Following up on a tangent to a SQLAlchemy issue starting [here](https://github.com/sqlalchemy/sqlalchemy/issues/5334#issuecomment-668179120) I have found that `fast_executemany` will cause a Python app to have continuously increasing memory consumption when it performs repeated `.executemany` calls involving `varchar(max)` columns. The same is not true when `varchar(100)` columns are involved.

Repro code:

``` python
from time import time
import psutil
import pyodbc

varchar_limit = "max"
print(f"using varchar({varchar_limit})")

print(f"pyodbc {pyodbc.version}")
cnxn = pyodbc.connect("DSN=mssqlLocal64", autocommit=True)
process = psutil.Process()


def print_status(msg, t0=None):
    s = f"{msg}: "
    mb = process.memory_info().vms / 1048576
    s += f"vms {mb:0.1f} MiB"
    if t0:
        s += f", {(time() - t0):0.1f} sec."
    print(s)


print_status("startup")
num_rows = 100_000
data = [(i, f"col{i:06}", 3.14159265 * (i + 1)) for i in range(num_rows)]
print_status("data loaded")

table_name = "pd_test"
col_names = ["id", "txt_col", "float_col"]
ins_sql = f"INSERT INTO {table_name} ({','.join(col_names)}) VALUES ({','.join('?' * len(col_names))})"
for iteration in range(5):
    t0 = time()
    crsr = cnxn.cursor()
    crsr.execute(f"DROP TABLE IF EXISTS {table_name}")
    crsr.execute(f"CREATE TABLE {table_name} (id int, txt_col varchar({varchar_limit}), float_col float(53))")
    crsr.fast_executemany = True
    crsr.executemany(ins_sql, data)
    crsr.close()
    print_status(f"iteration {iteration}", t0)
```

Console output:

``` none
using varchar(100)
pyodbc 4.0.30
startup: vms 8.7 MiB
data loaded: vms 31.1 MiB
iteration 0: vms 34.2 MiB, 53.8 sec.
[...]

PR #832, opened 2020-10-05
Bugfix/sql put data unicode to bytes memory leak

Mizaro:
This is a fix to prevent the memory leak from Issue #802 

In my investigation of the code's behavior, It seems that when there is a need to convert Unicode string to bytes, the newly created objects of bytes will never get dereference, thus staying in memory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,14 @@ pyodbc.conf
 # The Access unit tests copy empty.accdb and empty.mdb to these names and use them.
 test.accdb
 test.mdb
+
+# JetBrain's IDE
+.idea/
+
+# Visual Studio IDE
+.vs/
+*.sln
+*.vcxproj*
+
+# Visual Studio build folder
+x64/
--- a/src/params.cpp
+++ b/src/params.cpp
@@ -1987,6 +1987,11 @@ bool ExecuteMulti(Cursor* cur, PyObject* pSql, PyObject* paramArrayObj)
                         offset += remaining;
                     }
                     while (offset < cb);
+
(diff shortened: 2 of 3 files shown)


gordthompson:
@Mizaro - BTW, if you push another commit to this PR it may also fix the AppVeyor failure. I just submitted a trivial PR (#848)  as a test and its AppVeyor check passed. It looks like the AppVeyor CI environment has been changed since you submitted this PR.

gordthompson:
Hey @Mizaro - Any chance that you could apply the .gitignore patch in my previous comment? @mkleehammer might be more inclined to merge the PR if he doesn't have to mess around with conflicts. 

(I tried to submit the change as a PR to your fork, but GitHub wouldn't let me. I could submit PRs to a whole bunch of other forks, but yours wasn't among them. I've encountered the same issue in the past, but I don't understand why it happens.)

Mizaro:
@gordthompson  I will do it! Just not today 😅

mkleehammer:
Manually squashed and rebased.  Thanks!

Mizaro:
@mkleehammer, Ok, do you approve the change ?

Mizaro:
@mkleehammer, Thanks for inserting it :)

Mizaro:
@gordthompson Thanks for being so proactive !

gordthompson:
@thealbres - This patch was merged as 716572a1ca4426483bb31d39802b543d44ac6732 on 2021-01-21 and included in version 4.0.31, released 2021-07-03.

PR #855, opened 2021-01-22
Move Travis unit test code to the correct function

keitherskine:
As part of the merge of PR 773, some code in one of the unit tests ended up in a different unit test, causing the Travis CI tests to fail.  This PR fixes that.  The Travis CI tests should now pass.


mkleehammer:
You are fast!  I was about to dig into that.  Following the GitHub manual instructions is a bad idea and performs a rebase on the pull if you have --rebase as your default for pulls, which I do.  When I pulled one branch it went through about 30 commits and I hard to fix merge problems in about 10!  I'm quite sure I messed up the test.  Thanks.

PR #856, opened 2021-01-23
Fix timestamp fraction digits conversion for DecimalDigits larger than 10

cjalmeida:
For some reason, when using MS SQL Server ODBC Driver 17, against MS SQL Server 17, the `DecimalDigits` param info value is `10`, giving negative index in fast_pow10, thus making array access out of bounds.

In most situations, this would give us a garbage value and likely "truncate" the fraction. In my particular case, the garbage value was `0`, raising a division by zero FPE.

This PR is a quick fix, by adding a 10th element in fast_pow10. A more "robust" solution would be to add a "slow" version if `DecimalDigits > 10` as a fallback.

### DIFF
--- a/src/params.cpp
+++ b/src/params.cpp
@@ -403,9 +403,10 @@ static int PyToCType(Cursor *cur, unsigned char **outbuf, PyObject *cell, ParamI
         pts->second = PyDateTime_DATE_GET_SECOND(cell);
 
         // Truncate the fraction according to precision
-        long fast_pow10[] = {1,10,100,1000,10000,100000,1000000,10000000,100000000,1000000000, 10000000000};
+        size_t digits = min(9, pi->DecimalDigits);
+        long fast_pow10[] = {1,10,100,1000,10000,100000,1000000,10000000,100000000,1000000000};
         SQLUINTEGER milliseconds = PyDateTime_DATE_GET_MICROSECOND(cell) * 1000;
-        pts->fraction = milliseconds - (milliseconds % fast_pow10[10 - pi->DecimalDigits]);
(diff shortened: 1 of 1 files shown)


mkleehammer:
The SQL Server unit tests are failing.  I’m away from my computer and can’t look into it.  Can you look into that?   Also, addding a test that fails with the previous code would be helpful

cjalmeida:
Sorry taking the time, fixed the issue with proper truncation.

src/params.cpp
```diff
@@ -403,9 +403,10 @@ static int PyToCType(Cursor *cur, unsigned char **outbuf, PyObject *cell, ParamI
         pts->second = PyDateTime_DATE_GET_SECOND(cell);
 
         // Truncate the fraction according to precision
+        size_t digits = min(9, pi->DecimalDigits);
         long fast_pow10[] = {1,10,100,1000,10000,100000,1000000,10000000,100000000,1000000000};
         SQLUINTEGER milliseconds = PyDateTime_DATE_GET_MICROSECOND(cell) * 1000;
```
keitherskine, line 408: This looks like a perfectly reasonable solution.  Python datetimes can only handle microseconds anyway, so any more precision than that is essentially meaningless.

Whilst we're here though, is it worth changing the name of the variable "milliseconds" to "billionths" instead?  "Milliseconds" is a confusing misnomer.  When microseconds are multiplied by 1000, the result is billionths of a second, not milliseconds.

gordthompson: "nanoseconds" ref:

https://github.com/mkleehammer/pyodbc/issues/908#issuecomment-856871022

keitherskine: Nanoseconds!  Yes, not "billionths".  Much better, thank you Gord.

Issue #858, opened 2021-01-29
segfault on Linux  / unixODBC when calling "use" statement, master only; 4.0.30 is fine


zzzeek:
### Environment

- Python:  Python 3.8.3 (default, May 23 2020, 16:34:37) 
[GCC 9.3.1 20200408 (Red Hat 9.3.1-2)] on linux

- pyodbc:  4.0.31b51 from 7c7b1b1a928752158517f5c30165

- ODBC headers / devel libs (Fedora 33): 

    unixODBC-2.3.9-1.fc33.x86_64
    unixODBC-devel-2.3.9-1.fc33.x86_64

- OS:  
    uname -a
    Linux photon3 5.10.8-200.fc33.x86_64 #1 SMP Sun Jan 17 19:53:01 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

- DB:

    Centos packages:

    mssql-server-14.0.3356.20-23.x86_64
    mssql-tools-17.6.1.1-1.x86_64
    mssql-server-fts-14.0.3356.20-23.x86_64

- driver:  msodbcsql  13.1.9.2  

### Issue

SQLAlchemy's test suite needs to use the "use" statement during a setup process in order to create a schema in a certain database.  Building with latest master produces a segfault with the program below.  I can do a source build against the 4.0.30 tag and that one is fine.

```python
import pyodbc
print("Using pyodbc version: %s" % pyodbc.version)

conn = pyodbc.connect(
    "DRIVER={ODBC Driver 13 for SQL Server};Server=mssql2017,1433;Database=test;UID=scott;PWD=tiger^5HHH"
)
cursor = conn.cursor()

cursor.execute("select db_name()")
db_name = cursor.fetchone()[0]

cursor.execute("use %s" % db_name)
cursor.close()
conn.close()

print("OK!")
```

output:

```
Using pyodbc version: 4.0.31b51
Segmentation fault (core dumped)

```

v-chojas:
Please provide an ODBC trace. That will better show exactly where it faults.

Also note that latest version of msodbcsql is *17.7* , which has fixed several bugs since 13.x.

zzzeek:
right, forgot about those. 

When the trace is turned on, the output includes an additional message:

```
Using pyodbc version: 4.0.31b51
*** buffer overflow detected ***: terminated
Aborted (core dumped)

```

file is attached, seems like it just generally has more logging turned on.  

[...]

PR #859, opened 2021-01-31
Fix segmentation fault from GetDiagRecs()

keitherskine:
This PR should fix the segmentation fault described in #858 .  Function GetDiagRecs() should always return an integer otherwise the stack can become corrupted.  Still not sure why this impacted only RedHat and Oracle Linux distributions but it has to be fixed regardless.  I tested this on a RedHat 8.3 instance.

Whilst we're here, add one to the size of the MessageText string for the terminating NULL character.  Also, add more comments.


keitherskine:
@mkleehammer This PR is ready for review.

Issue #831, opened 2020-10-03
Type annotations for mypy


MartinThoma:
I've recently seen that pyodbc is one of the projects that doesn't come with type hints. It would be pretty awesome if one could add this.

I wanted to add them myself, but the protect seems to be almost purely C++. I have no clue how to properly annotate that. My hope was that we could add stub files (`.pyi`) for it. Does anybody have an idea how to do this? @mkleehammer Would you be open to add such stub files?

mkleehammer:
I'm certainly open to it, but like you I haven't researched it yet.  I'll look into it after I get this bug list down and make sure it is stable.

MartinThoma:
If you want general information about type annotations: https://medium.com/analytics-vidhya/type-annotations-in-python-3-8-3b401384403d

For libraries, there are 3 options:

* Full integration: I would consider that if you want to support Python 3.6+. If you want to support old Python versions, this will not work.
* Type comments: Directly in the code. IMHO it looks ugly, but serves as documentation as well.
* Type stubs: Extra files with type hints. You will only have the function signatures (replace the body by `...`) and store the file as `.pyi`. It does not have any effect on running Python, but still allows people/the editor to use the annotation.

keitherskine:
Fixed by #864.

jonaslb:
pyodbc is still not compatible with static checking with mypy / according to specifications due to issues https://github.com/python/mypy/issues/16987 / https://github.com/python/typing/issues/1333 . Essentially, there appears to be no proper way of annotating a top level single-file module. I have found no good workarounds besides a mypy override to ignore it.

PR #864, opened 2021-02-06
Add pyodbc.pyi stub file for type annotations and intellisense code completion

keitherskine:
This PR should address #831.

pyodbc does not currently have any type annotations for its classes and functions.  Type annotations are used by [mypy](https://mypy.readthedocs.io/en/stable/) and other linters to provide a measure of static type analysis for up-front error checking.  Also, pyodbc is a pure C++ extension module which means it's a Python module that doesn't contain any Python.  This makes it difficult for IDEs to provide "intellisense" code completion when developers are writing code that uses pyodbc.

By creating a `pyodbc.pyi` file, I'm hoping to address both of these issues.  I added the stub file to the `src` directory, and included it in `setup.py` as a [data file](https://docs.python.org/3/distutils/setupscript.html#installing-additional-files) so that it gets installed in the top-level pyodbc directory within the `site-packages` directory.  That way, IDEs should be able to find the stub file because it will be alongside the `pyodbc.py` file generated during installation (this is the preferred approach described in [PEP-0561](https://www.python.org/dev/peps/pep-0561/#stub-only-packages)).  Doing this, intellisense worked for me in VS Code on Windows with the Pylance extension, without the need for me to do anything apart from pip-installing pyodbc.  For example, intellisense:

image

and type annotation/checking:

image

Quick aside.  Strictly speaking pyodbc is not a Python package, it's just a Python module, so the packaging approach for distribution described in PEP-0561 does not apply here.  From PEP-0561 "This PEP does not support distributing typing information as part of module-only distributions.", hence the use of the "data_files" construct in `setup.py`.

Lastly, whilst in this neck o' the woods, I corrected the name of the "paramstyle" attribute in the pyodbc module doc description.


keitherskine:
Travis has errored out on one of the jobs with a "502 Bad Gateway" when pulling a docker image from Microsoft.  @mkleehammer , if you're able to simply restart that one job, I'm pretty sure if will succeed second time round.

keitherskine:
@mkleehammer This PR is ready for review.

alherm7:
I had an issue with the location of the .pyi file. When I installed pyodbc in my virtual environment the .pyi got placed into the top level folder of the virtual environment, i.e. "./.venv/" . Intellisense in VS Code was not able to find this location by default, so I had to manually move it into "./.venv/lib/python3.9/site-packages/" folder. This has also been described by others in issue #925 . Will this issue be fixed in future releases?

PR #924, opened 2021-07-06
Add 'closed' property to Connection

### LINKED ISSUE #65: Add property to check if a connection is open.

tekknolagi:
Fixes #65

Supersedes #70

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -968,6 +968,28 @@ static int Connection_setautocommit(PyObject* self, PyObject* value, void* closu
 }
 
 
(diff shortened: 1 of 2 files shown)


mkleehammer:
Looks good.  Thanks.

tekknolagi:
Awesome! Glad this came full circle. This may be the only thing that survived that internship :')

Issue #927, opened 2021-07-09
Wrong typings for the module


dfrison01:
### Environment

- Python: 3.7 x32
- pyodbc: 4.0.31
- OS: W10 x64

- Editor: VSCode, with Pylance enabled

### Issue

As illustrated in the attached images:
- The code inspector complains about pyodbc.connect() being used in a "with" statement.
- The code inspector complains about the utilization of the built-in "sorted()" with a list of Rows returned by the cursor.fetchall() method. The type of the "pyodbc.Row" object seems to be incorrect. As far as I know, the Row object should behave a lot like (and have the same methods as) a Sequence. This obviously happens with other functions, I'm only using sorted as an example.
- Both these warnings seems to be wrong, since the pyodbc module correctly handles the situations mentioned above. Anyway, the warnings are still very annoying.

I'm sorry if the description is poor since I don't have advanced knowledge in both the language and the module. I'm available to supply any other detail as needed.
Thanks a lot for the attention!

![pyodbc1](https://user-images.githubusercontent.com/9609651/125137687-93fd9680-e0e3-11eb-80fe-f94dd45fae25.JPG)
![pyodbc2](https://user-images.githubusercontent.com/9609651/125137695-995ae100-e0e3-11eb-93f1-2389ab7ced0e.JPG)
![pyodbc3](https://user-images.githubusercontent.com/9609651/125137705-9cee6800-e0e3-11eb-9994-45667ea297d3.JPG)

keitherskine:
Hi @dfrison01 ,
I'm having some difficulty reproducing your issue.  When I use Pylance and pylint to analyze a similar code snippet in VS Code, I get the following:
![image](https://user-images.githubusercontent.com/1158224/125176234-8ace0180-e197-11eb-915b-7fcb94a4ac37.png)
I do get an error of sorts on `pyodbc.connect()` but no issue with the `__enter__` function.  And I don't get any error on the `sorted()` function.

Are you perhaps using a linter different from pylint?  Every linter has its own quirks.

It may help to look at the logs for Pylance.  In VS Code, you can see those logs by opening up the terminal, selecting "OUTPUT" and choosing "Python Language Server" in the dropdown.  You should see something like this:
![image](https://user-images.githubusercontent.com/1158224/125176340-3d9e5f80-e198-11eb-8203-827d35ddacd4.png)
You may need to set the log level to "Trace" to see more information in the logs.  Here's part of my VS Code settings file:
```
[...]

PR #932, opened 2021-07-17
Fixes to pyodbc.pyi for #927

keitherskine:
This PR addresses some issues with the pyodbc.pyi file spelled out in #927 .  It also makes some other cosmetic tweaks to the file.  Specifically:

- Add dunder methods (e.g. `__enter__`) to the classes to make them explicit.  Pylance seems to prefer it that way.
- Add defaults to a couple of the parameters for the `connect()` function.  Again, Pylance seems to need them.
- Remove the `_TV` object which seemed like a good idea at the time but actually just confuses matters.
- Other cosmetic changes: re-arranging some of the constants to make them more readable; adding/updating comments.

Hopefully this will make pyodbc.pyi more compatible with Pylance, mypy, and other static typecheckers.  Note, this PR does not address the issue (925) of how pyodbc.pyi is added to the distributed package.

### DIFF
--- a/src/pyodbc.pyi
+++ b/src/pyodbc.pyi
@@ -1,38 +1,38 @@
 from __future__ import annotations
-from typing import Any, Callable, Dict, Generator, Iterator, List, Optional, Sequence, Tuple, TypeVar, Union
-
-_TV = TypeVar('_TV')  # represents any class, for the cursor description "type_code" attribute
+from typing import Any, Callable, Dict, Generator, Iterator, List, Optional, Sequence, Tuple, Union
 
 
-# ODBC connection attributes
-SQL_ACCESS_MODE: int
-SQL_AUTOCOMMIT: int
-SQL_LOGIN_TIMEOUT: int
-SQL_OPT_TRACE: int
-SQL_OPT_TRACEFILE: int
-SQL_TRANSLATE_DLL: int
-SQL_TRANSLATE_OPTION: int
-SQL_TXN_ISOLATION: int
-SQL_CURRENT_QUALIFIER: int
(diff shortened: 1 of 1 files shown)


src/pyodbc.pyi
```diff
@@ -404,11 +442,11 @@ def getDecimalSeparator() -> str: ...
...
-            /, *,  # positional-only parameters before, named-only after
+            /, *,  # only positional parameters before, only named parameters after
             autocommit: bool = False,
             encoding: str = 'utf-16le',
             ansi: bool = False,
             readonly: bool = False,
-            timeout: int,
-            attrs_before: dict,
+            timeout: int = 0,
+            attrs_before: dict = {},
```
imba-tjd, line 451: Hi. I have little experience in pyodbc, but typically things like `{}` or `[]` are not good for default values. Are you sure it's correct?

v-chojas: I don't know what you mean by 'not good for default values", but the default is empty so it is correct..

imba-tjd: See https://github.com/satwikkansal/wtfpython#-beware-of-default-mutable-arguments

keitherskine: You're certainly correct that real code should not typically have mutable values as default parameter values (because default values are created only once, not every time the function is called).  This code however is in the .pyi file which is used only for typing.  If these two defaults are left out of the function definition, it looks like the two parameters are mandatory which is not the case.

v-chojas: Also, these functions are implemented in native code.

PR #951, opened 2021-09-08
Fix for DSN Names with non-ASCII chars

v-makouz:


keitherskine:
With respect to the AppVeyor build failures, it would appear that support for Python 3.5 has been dropped from pip 21.0.  I'm guessing this is why the `--no-warn-script-location` option is not being recognized in the Python 3.5 builds.

From `setup.py`, pyodbc currently supports Python 3.5 thru 3.8.  There's a good case for dropping 3.5 and adding 3.9.  Let me see if the build scripts work with 3.9.

PR #952, opened 2021-09-12
Add support for Python 3.10, drop EOL 3.5

hugovk:
Fixes https://github.com/mkleehammer/pyodbc/issues/954.

Python 3.10.0 final is due for release in October:

* https://www.python.org/dev/peps/pep-0619/

The final release candidate is now out and the Python release team has issued a call to action for community members:

> We strongly encourage maintainers of third-party Python projects to prepare their projects for 3.10 compatibilities during this phase. As always, report any issues to the Python bug tracker.

https://discuss.python.org/t/python-3-10-0rc2-is-now-available/10496?u=hugovk

So let's also test and add support for 3.10.

Drops support for EOL Python 3.5.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -110,6 +110,8 @@ def main():
                        'Programming Language :: Python :: 3.6',
                        'Programming Language :: Python :: 3.7',
                        'Programming Language :: Python :: 3.8',
+                       'Programming Language :: Python :: 3.9',
+                       'Programming Language :: Python :: 3.10',
                        'Topic :: Database',
                        ],
 
--- a/setup.py
+++ b/setup.py
@@ -106,7 +106,6 @@ def main():
                        'Programming Language :: Python :: 2',
                        'Programming Language :: Python :: 2.7',
                        'Programming Language :: Python :: 3',
-                       'Programming Language :: Python :: 3.5',
                        'Programming Language :: Python :: 3.6',
                        'Programming Language :: Python :: 3.7',
                        'Programming Language :: Python :: 3.8',
--- a/setup.py
+++ b/setup.py
@@ -96,6 +96,8 @@ def main():
 
         'license': 'MIT',
 
+        'python_requires': '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*',
+
         'classifiers': ['Development Status :: 5 - Production/Stable',
                        'Intended Audience :: Developers',
                        'Intended Audience :: System Administrators',
--- a/.github/workflows/ubuntu_build.yml
+++ b/.github/workflows/ubuntu_build.yml
@@ -11,17 +11,17 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - python-version: 2.7
+          - python-version: "2.7"
             tests-dir: tests2
-          - python-version: 3.6
+          - python-version: "3.6"
             tests-dir: tests3
-          - python-version: 3.7
+          - python-version: "3.7"
             tests-dir: tests3
-          - python-version: 3.8
+          - python-version: "3.8"
             tests-dir: tests3
-          - python-version: 3.9
+          - python-version: "3.9"
             tests-dir: tests3
-          - python-version: 3.10-dev
+          - python-version: "3.10"
             tests-dir: tests3
 
     services:
--- a/.github/workflows/ubuntu_build.yml
+++ b/.github/workflows/ubuntu_build.yml
@@ -13,8 +13,6 @@ jobs:
         include:
           - python-version: 2.7
             tests-dir: tests2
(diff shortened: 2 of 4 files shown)


keitherskine:
@hugovk Many thanks for this PR, but could you update `setup.py` to reflect the addition of Python 3.9 and 3.10?

Also, the build tests are failing for Python 3.5.  Rather than fix the unit tests, it seems easiest to add #954 into this PR.  If you're OK with that, go ahead and comment out the relevant Python 3.5 tests in `appveyor.yml`, and `ubuntu_build.yml`, add Python 3.9 tests to `appveyor.yml` (3.10 is not available there), and remove the reference to Python 3.5 in `setup.py`.  Then the build tests for both Windows and Ubuntu should pass.  Many thanks!

hugovk:
> @hugovk Many thanks for this PR, but could you update `setup.py` to reflect the addition of Python 3.9 and 3.10?

I already added the 3.9 and 3.10 classifiers to `setup.py` 👍 

> Also, the build tests are failing for Python 3.5. Rather than fix the unit tests, it seems easiest to add #954 into this PR. If you're OK with that, go ahead and comment out the relevant Python 3.5 tests in `appveyor.yml`, and `ubuntu_build.yml`, add Python 3.9 tests to `appveyor.yml` (3.10 is not available there), and remove the reference to Python 3.5 in `setup.py`. Then the build tests for both Windows and Ubuntu should pass. Many thanks!

Good idea, removed 3.5!

I also added `python_requires` to `setup.py`, so pip knows which package version to install for end users.

---

Whilst we're at it, how about also dropping 2.7? It's been EOL for nearly two years and is little used. Dropping would further reduce the maintenance burden and allow modern features of Python to be used.

Here's the pip installs for pyodbc from PyPI for August 2021:

| category | percent | downloads  |
| -------- | ------: | ---------: |
| 3.7      |  71.38% |  8,562,696 |
| 3.8      |  15.27% |  1,831,650 |
| 3.6      |   6.20% |    743,452 |
| 2.7      |   3.02% |    362,048 |
| 3.5      |   2.73% |    327,753 |
| 3.9      |   0.91% |    109,209 |
| null     |   0.49% |     59,002 |
| 3.4      |   0.00% |        576 |
| 3.10     |   0.00% |        330 |
| 3.3      |   0.00% |          3 |
| 2.6      |   0.00% |          1 |
| Total    |         | 11,996,720 |

Source: `pip install -U pypistats && pypistats python_minor pyodbc --last-month`

hugovk:
Updated now Python 3.10 final is out.

https://discuss.python.org/t/python-3-10-0-is-now-available/10955

keitherskine:
@hugovk AppVeyor have at last added Python 3.10 to their Windows VS2019 build.  In which case, it would be great if you could add Python 3.10 to the pyodbc `appveyor.yml`, if you get a chance.  Many thanks!

For ref: https://www.appveyor.com/docs/windows-images-software/#python

setup.py
```diff
@@ -96,6 +96,8 @@ def main():
 
         'license': 'MIT',
 
+        'python_requires': '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*',
```
keitherskine, line 99: As I understand it, this "python_requires" statement would prevent pip (or at least recent versions of pip) from installing pyodbc on Python 3.5, 3.4, etc.  Do we really want to go that far?  It would be unusual (and not recommended) to install pyodbc on Python 3.4 but do we want to positively forbid it?

abitrolly: Bump major version then?

hugovk: `python_requires` doesn't positively forbid installation, people can tell pip to ignore it.

```
  --ignore-requires-python    Ignore the Requires-Python information.
```

`python_requires` gives protection so you can use modern syntax without worrying you're going to break things for end users. And if you're not testing a version, you shouldn't give the impression it's supported.

On the whole, it's not just worth having half-in/half-out support, especially as old versions are so little used.

On the topic: we could also upgrade to modern Python syntax now (f-strings!). I could do it here or in a followup PR?

hugovk:
> @hugovk AppVeyor have at last added Python 3.10 to their Windows VS2019 build. In which case, it would be great if you could add Python 3.10 to the pyodbc `appveyor.yml`, if you get a chance. Many thanks!
> 
> For ref: [appveyor.com/docs/windows-images-software/#python](https://www.appveyor.com/docs/windows-images-software/#python)

Done!

gordthompson (approved):
Looks good to me. Would be nice to get CI working again.

keitherskine (approved):

hugovk:
By the way, Python 3.6 is EOL in a couple of weeks :)

https://endoflife.date/python

PR #953, opened 2021-09-12
Replace deprecated SafeConfigParser with ConfigParser

hugovk:
Replace deprecated `SafeConfigParser` with `ConfigParser`, deprecated since Python 3.2 and potentially being removed in Python 3.11 (due for release October 2022).

* https://bugs.python.org/issue45173

Remove some unused imports from the same file.

### DIFF
--- a/tests3/testutils.py
+++ b/tests3/testutils.py
@@ -1,5 +1,5 @@
 import os, sys, platform
-from os.path import join, dirname, abspath, basename
+from os.path import join, dirname, abspath
 import unittest
 from distutils.util import get_platform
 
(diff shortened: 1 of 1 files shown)


keitherskine (approved):
There are similar `SafeConfigParser` references in `tests2/testutils.py`, but I don't think we need to update any Python 2 code at this point, unless absolutely necessary.

keitherskine:
Many thanks for this, @hugovk .  Good to keep on top of this.

Of course, the build checks are failing for Python 3.5, as noted in #952, but hopefully #952 should fix that.

hugovk:
Rebased now https://github.com/mkleehammer/pyodbc/pull/952 is merged, and all green! ✅

PR #987, opened 2021-12-07
Designate connection string argument type as optional

ajnelson-nist:
The signature for `pyodbc.connect()` designates the connection string as
required. However, the implementation in `src/pyodbcmodule.cpp`
`mod_connect` permits the string to be absent if there are no positional
arguments, relying instead on only keyword arguments.  The doc string in
`src/pyodbcmodule.cpp` `connect_doc` also confirms the string may be
optionally supplied.

This patch updates the type signature to reflect the implementation and
documentation.

Signed-off-by: Alex Nelson <alexander.nelson@nist.gov>


ajnelson-nist:
Thank you!

PR #992, opened 2021-12-18
Solving issue with infinity dates

lsrzj:
PostgreSQL ODBC driver returns year -9999 for -infinity dates, so python raises ValueError exception when reading cursor data, because the year field cannot be less than 1. So, I made modifications to bring the value brought by the ODBC driver to python's accepted limits that are years between 1 and 9999. That way the cursor will not raise any exception and pyodbc becomes compatible with infinity dates, just like psycopg2. The big changes seen on removed and added lines are due to autoformat tool in my VSCode, the only part I touched was the switch case part, to add the SQL_TYPE_TIMESTAMP to correct the values to python's limits.

### DIFF
--- a/src/getdata.cpp
+++ b/src/getdata.cpp
@@ -36,7 +36,7 @@ void GetData_init()
     PyDateTime_IMPORT;
 }
 
-static byte *ReallocOrFreeBuffer(byte *pb, Py_ssize_t cbNeed);
+static byte* ReallocOrFreeBuffer(byte* pb, Py_ssize_t cbNeed);
 
 inline bool IsBinaryType(SQLSMALLINT sqltype)
 {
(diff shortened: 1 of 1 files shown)


lsrzj:
The big changes seen on removed and added lines are due to autoformat tool in my VSCode, the only part I touched was the switch case part, to add the SQL_TYPE_TIMESTAMP to correct the values to python's limits.

mkleehammer:
Thanks for this.

Can we get a copy *without* the autoformatting, though?  It makes it hard to diff things over time.

lsrzj:
> Thanks for this.
> 
> Can we get a copy _without_ the autoformatting, though? It makes it hard to diff things over time.

For sure, make it the best way for the project. I forgot to disable the auto format tool.

PR #999, opened 2022-01-02
Replace deprecated PyUnicode_FromUnicode(NULL, size) calls (#998)

### LINKED ISSUE #998: Deprecation warning breaks CGI scripts
### Environment

- Python: 3.10.1
- pyodbc: 4.0.32
- OS: Any
- DB: Any
- driver: Any

### Issue

Under Python 3.10 `PyUnicode_FromUnicode(NULL, size)` triggers a `DeprecationWarning` message to `stderr`, breaking CGI scripts. To reproduce:
```python
#!/usr/bin/env python3

"""Repro case to demonstrate deprecation warning.

Breaks CGI scripts.
"""

from argparse import ArgumentParser
from pyodbc import connect  # pylint: disable=no-name-in-module

parser = ArgumentParser()
parser.add_argument("--connection-string", "-c", required=True)
opts = parser.parse_args()
conn = connect(opts.connection_string)
cursor = conn.cursor()
cursor.execute("SELECT 1 AS a, 2 AS b")
rows = cursor.fetchall()
print(f"""\
Content-type: text/plain

HTTP header(s) should be first in the output, but they're not. :(
{rows!r}""")
```
#### Expected Output
```
Content-type: text/plain

HTTP header(s) should be first in the output, but they're not. :(
[(1, 2)]
```
#### Actual Output
```
/Users/bkline/gdrive/Notes/Python/./pyodbcrepro.py:18: DeprecationWarning: PyUnicode_FromUnicode(NULL, size) is deprecated; use PyUnicode_New() instead
  print(f"""\
Content-type: text/plain

HTTP header(s) should be first in the output, but they're not. :(
[(1, 2)]
```

bkline:
Current versions of Python write a deprecation warning message to stderr, which breaks CGI scripts running under web servers which fold stderr into stdout. Likely breaks other software. This change replaces the deprecated calls with PyUnicode_New(size, 0x10ffff). The second argument represents the highest code point for Unicode characters, which is used because the calls being replaced have no string value from which we can determine the smallest number of bytes needed to represent any value which can be stored in the newly-created object.

Fixes #998

### DIFF
--- a/src/pyodbccompat.h
+++ b/src/pyodbccompat.h
@@ -66,13 +66,17 @@ inline void PyString_ConcatAndDel(PyObject** lhs, PyObject* rhs)
 
 #endif
 
+#ifndef MAX_UNICODE
+  #define MAX_UNICODE 0x10ffff
+#endif
+
 inline PyObject* Text_New(Py_ssize_t length)
 {
     // Returns a new, uninitialized String (Python 2) or Unicode object (Python 3) object.
 #if PY_MAJOR_VERSION < 3
     return PyString_FromStringAndSize(0, length);
 #else
-    return PyUnicode_FromUnicode(0, length);
+    return PyUnicode_New(length, MAX_UNICODE);
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


bkline:
I have included a unit test, but I assume others more familiar with the project will want to decide how to integrate it with the existing tests. The test fails without the fix and passes with it.

bkline:
Reaching out in the [ticket](https://github.com/mkleehammer/pyodbc/issues/998#issuecomment-1003718340) for assistance.

bkline:
Going to get another running start on this one.

Issue #945, opened 2021-08-22
Freeze Python 2.7 and modernize codebase


mkleehammer:
I'd like to look into splitting the codebase into Python 2 and Python 3 and modernizing the 3 version.  There used to be a *huge* number of Python 2 downloads every day, but Python 3 is now ~95% of the pypi download traffic.

https://pypistats.org/packages/pyodbc

Does anyone know the best way to do this?  My first inclination is to make a new v5 that is only the modern version and maintain the 4.x branch for security fixes.  Any other suggestions?

Some things I'm thinking about for a v5:

- Update C API and consider using the stable ABI.  If I understand correctly, it would allow new versions of Python to use the same binaries.
- Would it be helpful to have a Python layer that loads the C implementation as _pyodbc?  Does this give the project more flexibility to introduce items in whichever layer is easiest?
- More thoroghly integrate the parameter binding of per-value and fast_executemany so there is only 1 function for encoding each.
- Provide a connection factory to replace the hidden CnxnInfo.  It would be eaiser to configure pre-connect items on the factory and then ask for the connect.  (We might need an optional password on the connect for those that don't want it stored globally.)
- Switch to pytest
-

hugovk:
Many projects make/tag a last release for Python 2.7, then bump the major version and move on.

2.7 has been EOL for 20 months now so people using it aren't getting security releases, and most of the top-downloaded PyPI packages [have already dropped it](https://hugovk.github.io/drop-python/2.7/
) (pip, setuptools, certifi, numpy, cryptography, pytest...).

Make sure you have `python_requires>=3.6` in the new version, then pip won't install the new version for folk on Python 2.7.

keitherskine:
It's been a while since @mkleehammer raised this issue but I think everything that @mkleehammer mentioned in his original post are great ideas.  Dropping Python 2.7 support makes perfect sense and will make the codebase much easier to maintain.  As I understand it, we more-or-less have to start using the [stable API](https://docs.python.org/3/c-api/stable.html#stable) for Unicode strings soon because Python 3.12 won't work with pyodbc in its current form (there are already a bunch of deprecation warnings during builds).  I also like the idea of putting the C code into a `_pyodbc` module and creating a wrapper Python `pyodbc` module that calls it.  This would make the pyodbc API easier to understand, and make code changes easier too.

[...]

Issue #917, opened 2021-06-20
Deprecated PyUnicode functions that must be updated


keitherskine:
I just wanted to create this issue as a placeholder for certain c-extension functions used by pyodbc that are going to be removed in upcoming Python versions.  Here is the list I have so far:

To be removed in version 3.11:
PyUnicode_EncodeUTF8
PyUnicode_EncodeUTF16

To be removed in version 3.12:
PyUnicode_GET_SIZE
PyUnicode_AS_UNICODE
PyUnicode_AS_DATA
PyUnicode_FromStringAndSize
PyUnicode_FromUnicode
PyUnicode_GetSize
PyUnicode_FromObject

Ref:
https://docs.python.org/3/c-api/unicode.html
https://docs.python.org/3/c-api/call.html

keitherskine:
There is one PR #792 that has made a start on converting the relevant code to the preferred functions.

mkleehammer:
I saw the PR - great job.  I left a question on one part.

This may be the first item for v5 *(#945).

bkline:
@mkleehammer Under Python 3.10 `PyUnicode_FromUnicode(NULL, size)` triggers a `DeprecationWarning` message to `stderr`, breaking CGI scripts. Should I open a separate ticket to get this fixed on a faster track than #945?
To reproduce (`pyodbc-4.0.32`):
```python
...
rows = cursor.fetchall()
logstring = repr(rows)
```
(Also, GitHub is somehow suppressing the comment you said you left on #792.)

gordthompson:
What's the status of this now that #792 has been merged?

keitherskine:
There are still plenty of deprecated functions that are causing warning messages during builds.  Until the Python 3.12 deprecated functions are replaced in the code, this issue still stands.

opoplawski:
Fedora rawhide now has Python 3.12 and so pyodbc is failing to build.  Any progress on this issue?

mkleehammer:
Well, the py3 branch builds on Python 3.12 on Ubuntu, without fast_execute_many, and works with the PostgreSQL unit tests.

My conversion task list of this topic has a bunch of unchecked items, but I actually think they are done.  One difficultly with any PRs, is that the py3 branch is quite different from the production branch for now.

mkleehammer:
I've merged 5.0 with Python 3.12 support into the master branch and published 5.0.0 alpha2.  I'll make a final release of 5.0.0 in the next few days if no serious problems are found.

[...]

PR #1002, opened 2022-01-02
Replace deprecated PyUnicode_FromUnicode(NULL, size) calls (#998)

### LINKED ISSUE #998: Deprecation warning breaks CGI scripts

bkline:
Current versions of Python write a deprecation warning message to `stderr`, which breaks CGI scripts running under web servers which fold `stderr` into `stdout`, and likely breaks other software. This change replaces the deprecated calls with `PyUnicode_New(size, maxchar)`. The accompanying code to populate the new objects has also been rewritten to use the new PyUnicode APIs.

I have included a unit test, but I assume others more familiar with the project will want to decide how to integrate it with the existing tests. I didn't fold the test in with any of the other sets, because the only sets in tests3 which were not DBMS-specific were for tests of the API, and this isn't a fix of API behavior (or rather, it's a fix of internals related to working with the Python APIs). The test fails without the fix and passes with it.

This is my second shot at a PR for this issue. My first attempt took the instructions in the deprecation warning and in the documentation for `PyUnicode_New()` at face value, but that wasn't sufficient, and the pipeline tests didn't all pass. So I dug deeper into the [documentation for all of the PyUnicode APIs](https://docs.python.org/3/c-api/unicode.html) and [PEP 393](https://www.python.org/dev/peps/pep-0393/) and did some more extensive rewriting of the two parts of `pyodbc` which were calling `PyUnicode_FromUnicode(NULL, size)`. Assuming this second attempt passes the pipeline tests, I would very much like some very close scrutiny of this patch. I am very confident that my experience with the Python C APIs is much less extensive than the seasoned maintainers for the project, and the documentation is thinner than I would have liked, and downright buggy in places (for example, the documentation [claims](http://ld2015.scusa.lsu.edu/python-3.3.2-docs-html/c-api/unicode.html#PyUnicode_CopyCharacters) that `PyUnicode_CopyCharacters()` returns `0` on success, but I figured out from digging through the Python API's own C code that the function actually returns the number of characters copied; note also the mixup in the argument names—the second `to` should be `from`).

```
int PyUnicode_CopyCharacters(PyObject *to, Py_ssize_t to_start, PyObject *to, Py_ssize_t from_start, Py_ssize_t how_many)
```

> Copy characters from one Unicode object into another. This function performs character conversion when necessary and falls back to memcpy() if possible. Returns -1 and sets an exception on error, otherwise returns 0.

Fixes #998

### DIFF
--- a/src/pyodbcmodule.cpp
+++ b/src/pyodbcmodule.cpp
@@ -1327,7 +1327,6 @@ BOOL WINAPI DllMain(
 }
 #endif
 
-
 static PyObject* MakeConnectionString(PyObject* existing, PyObject* parts)
 {
     // Creates a connection string from an optional existing connection string plus a dictionary of keyword value
@@ -1344,13 +1343,13 @@ static PyObject* MakeConnectionString(PyObject* existing, PyObject* parts)
 
     I(PyUnicode_Check(existing));
 
-    Py_ssize_t length = 0;      // length in *characters*
-    if (existing)
-        length = Text_Size(existing) + 1; // + 1 to add a trailing semicolon
-
     Py_ssize_t pos = 0;
     PyObject* key = 0;
     PyObject* value = 0;
+    Py_ssize_t length = 0;      // length in *characters*
+#if PY_MAJOR_VERSION < 3
+    if (existing)
+        length = Text_Size(existing) + 1; // + 1 to add a trailing semicolon
 
     while (PyDict_Next(parts, &pos, &key, &value))
     {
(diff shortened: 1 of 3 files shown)


gordthompson:
With increasing adoption of Python 3.10 this will affect more and more people.

@mkleehammer , @v-chojas - Care to have a look?

mkleehammer:
Wow, lots of great work.  Thanks.  I'll spend some time looking through this carefully.

mkleehammer:
I wonder - as nice as this is, would it be better to hurry the v5?  A big part of the complexity is dealing with multiple encodings due to 2.7.  Python itself can deal with merging the different encodings into a single string trivially when using `''.join(a)`.  We could rely on Python more if we had a single version and possibly if we had the core in Python instead of C.

Thoughts?

bkline:
> I wonder - as nice as this is, would it be better to hurry the v5?

Well, sure. I only opened #998 (and the PRs) because I had never gotten a response to the question I asked last year in #917.

> Should I open a separate ticket to get this fixed on a faster track than https://github.com/mkleehammer/pyodbc/issues/945?

Not complaining. We're all busy. 😉

eabase:
Any reason why this PR is not yet merged? 

Is it possible this is also breaking the output of Pandas output when read into a DF from fetching an SQL result? 

I'm referring to these SO questions:
* https://stackoverflow.com/questions/71067094/pyodbc-deprecation-warning-when-printing
* https://stackoverflow.com/questions/72123992/panda-df-not-showing-all-rows-after-loading-from-ms-sql#

alanblyth:
This is a blocker for me, would be great to see this merged.

Issue #1004, opened 2022-01-04
pgtests test_emoticons_as_literal fails


bkline:
### Environment

- Python: 3.10.1
- pyodbc: built from current HEAD of master branch
- OS: macOS 11.6.2 (for client) Debian 10.2.1-6 (for server)
- DB: PostgreSQL 14.1
- driver: psqlodbcw.so 13.02.0000 (supports ODBC version 03.51)
- unixODBC 2.3.9

### Issue

The `test_emoticons_as_literal()` test fails for the `pgtests.py` suite.
```
======================================================================
FAIL: test_emoticons_as_literal (__main__.PGTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/bkline/repos/pyodbc/tests3/pgtests.py", line 625, in test_emoticons_as_literal
    self.assertEqual(result, v)
AssertionError: 'x � z' != 'x 🌜 z'
- x � z
?   ^
+ x 🌜 z
?   ^
```
This is similar to the failure reported for another test suite by https://github.com/mkleehammer/pyodbc/issues/536. Interestingly, if I run the tests with the `--ansi` flag set, this test passes, but the `test_output_conversion` test fails, because apparently setting `ansi=True` in the `pyodbc.connect()` call results for some reason in the return value from the local `convert()` function not being used (or perhaps that local function isn't invoked at all)
```
======================================================================
FAIL: test_output_conversion (__main__.PGTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/bkline/repos/pyodbc/tests3/pgtests.py", line 671, in test_output_conversion
    self.assertEqual(value, 'X123.45X')
AssertionError: '123.45' != 'X123.45X'
- 123.45
+ X123.45X
? +      +
```
However, if I comment out the `setencoding()` call:
```diff
diff --git a/tests3/pgtests.py b/tests3/pgtests.py
index dcb03a9..5dfacc9 100755
--- a/tests3/pgtests.py
+++ b/tests3/pgtests.py
@@ -70,7 +70,7 @@ class PGTestCase(unittest.TestCase):
 
         # I've set my test database to use UTF-8 which seems most popular.
         self.cnxn.setdecoding(pyodbc.SQL_WCHAR, encoding='utf-8')
-        self.cnxn.setencoding(encoding='utf-8')
+        # self.cnxn.setencoding(encoding='utf-8')
 
[...]

PR #1005, opened 2022-01-06
Disable setencoding/setdecoding in tests3/pgtests.py

gordthompson:
Fixes: #1004

Thanks to @bkline for reporting and diagnostics!


mkleehammer:
I am surprised.  Thanks.  Can I get you to delete the lines instead of commenting them out?  I think it would be cleaner for future devs.

mkleehammer:
Nevermind.  I'll take it as-is.  We'll be cleaning up tests a lot in v5.

PR #1013, opened 2022-01-20
Added InterfaceError to pyodbc.pyi.

bdholder:
Fixes: #1016


keitherskine:
Many thanks for this, @bdholder , it fixes an oversight.

Issue #1082, opened 2022-07-19
4.0.34 manylinux wheels include old libodbc.so and libltdl.so libraries


gordthompson:
### Environment

- Python: 3.8
- pyodbc: 4.0.34
- OS: Xubuntu 20.04
- DB: n/a
- driver: n/a

### Issue

With 4.0.34 we started building manylinux wheel files. Unfortunately, they include rather old versions of libodbc.so and libltdl.so that override the system versions, leading to errors as described in #1079 and #1081.

#### pyodbc 4.0.34

```
(pyodbc_test) gord@vbox-Xubu-20-04-a:~$ pip install pyodbc
Collecting pyodbc
  Downloading pyodbc-4.0.34-cp38-cp38-manylinux_2_24_x86_64.whl (475 kB)
  …
Successfully installed pyodbc-4.0.34
(pyodbc_test) gord@vbox-Xubu-20-04-a:~$ python
Python 3.8.10 (default, Jun 22 2022, 20:18:18) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyodbc
>>> pyodbc.__file__
'/home/gord/.virtualenvs/pyodbc_test/lib/python3.8/site-packages/pyodbc.cpython-38-x86_64-linux-gnu.so'
>>> quit()
(pyodbc_test) gord@vbox-Xubu-20-04-a:~$ ldd /home/gord/.virtualenvs/pyodbc_test/lib/python3.8/site-packages/pyodbc.cpython-38-x86_64-linux-gnu.so
	linux-vdso.so.1 (0x00007ffda9f7a000)
	libodbc-2003e41d.so.2.0.0 => /home/gord/.virtualenvs/pyodbc_test/lib/python3.8/site-packages/pyodbc.libs/libodbc-2003e41d.so.2.0.0 (0x00007fc4d0931000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fc4d073d000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fc4d05ee000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fc4d05d3000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fc4d05b0000)
[...]

PR #1084, opened 2022-07-20
Fix for #1082 libraries in Linux wheels

keitherskine:
It appears that the libodbc.so and libltdl.so libraries are added to the linux build wheels by the "repair" step in Github Actions.  This step adds any libraries it deems necessary to fulfil the "libraries" attribute in setup.py and also renames the linux build wheel files from "linux" to "manylinux".  As far as I can tell, that last rename is not strictly necessary, and can't be separated out, so I'm disabling that entire repair step.

Also, a few other notes:

- now using the latest version of cibuildwheel, 2.8.1
- for now, continue to use Debian-based linux images (rather than the latest AlmaLinux image)
- macos-10.15 will be removed on Aug 30, so update that
- don't build for musllinux, which seems to be problematic
- still not building 32-bit wheels, but we might want to re-visit that
- Debian9 installs unixODBC 2.3.4, which is old but I don't believe that is linked into the final wheels
- macOS wheels include the libodbc.2.dylib and libltdl.7.dylib libraries, which don't appear to be problematic, and libodbc seems to refer to unixODBC version 2.3.11 (the latest version, which is good)

Please do try out these new build wheels when you get a chance.

### DIFF
--- a/.github/workflows/ubuntu_build.yml
+++ b/.github/workflows/ubuntu_build.yml
@@ -197,18 +197,22 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, windows-2019, macos-10.15]
+        os: [ubuntu-20.04, windows-2019, macos-11]
 
     steps:
       - uses: actions/checkout@v2
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.1.3
+        uses: pypa/cibuildwheel@v2.8.1
         env:
+          # use the Debian9-based image manylinux_2_24 (https://github.com/pypa/manylinux)
           CIBW_MANYLINUX_X86_64_IMAGE: manylinux_2_24
+          # need unixODBC on linux
           CIBW_BEFORE_ALL_LINUX: apt-get update && apt-get -y install unixodbc-dev
-          # disable 32-bit and pypy builds
-          CIBW_SKIP: "*-win32 *-manylinux_i686 pp*"
+          # repairing the linux wheels adds old buggy versions of libraries to them, so don't (see issue 1081)
+          CIBW_REPAIR_WHEEL_COMMAND_LINUX: ""
+          # for now, disable 32-bit, musllinux, and pypy builds
+          CIBW_SKIP: "*-win32 *-manylinux_i686 *-musllinux_* pp*"
 
       - uses: actions/upload-artifact@v2
         with:
--- a/.github/workflows/ubuntu_build.yml
+++ b/.github/workflows/ubuntu_build.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   run_tests:
-
+    name: Run tests on Python ${{ matrix.python-version }}
     runs-on: ubuntu-latest
 
     strategy:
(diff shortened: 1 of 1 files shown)


v-chojas:
> macOS wheels include the libodbc.2.dylib and libltdl.7.dylib libraries, which don't appear to be problematic

I don't think that should be done either. While Mac doesn't have unixODBC by default, someone who has it (e.g. after installing msodbcsql) will encounter much confusion when pyodbc does something different than when the same connection is made in isql, and some ODBC drivers that call back into the DM will attempt to use the system one, which might not match. Suffice to say that multiple DM libs have caused a lot of problems in the past.

gordthompson:
I just checked a couple of the Linux wheels in here

https://github.com/mkleehammer/pyodbc/suites/7440010116/artifacts/304197945

and they look good to me. I do understand why some people would "like" to avoid installing unixODBC themselves, but they are simply unaware of the potential problems. Whatever the outcome—and I'm +1 on this approach—we'll need to update the "Installing on Linux" parts of the wiki, so we should remember to provide a brief explanation of why it's the way it is.

As for 32-bit Windows wheels, I would have been very keen on them a few years ago when I was doing a fair bit of MS Access stuff and Office installs were still mostly 32-bit. However, neither of those things is true now, and 32-bit wheels are still available at

https://www.lfd.uci.edu/~gohlke/pythonlibs/#pyodbc

so I would be inclined to leave them out for the time being.

gordthompson (approved):
I have tested the Linux wheels from the "artifacts" file at

https://github.com/mkleehammer/pyodbc/actions/runs/2704829437

against

- Alpine Linux
- Oracle Linux 8 and 9 (as a proxy for RHEL), and
- a couple of versions of Ubuntu

and they seem to work fine. Of course, unixODBC needed to be installed via

- `apk add unixodbc`
- `sudo yum install unixODBC`, and
- `sudo apt install unixodbc`

respectively.

I did notice that the macosx wheels also bundle the two .dylib files

```
pyodbc.dylibs/libtdl.7.dylib
pyodbc.dylibs/libodbc.2.dylib
```

but I have no way of testing them. Besides, they may have been built on something newer than the platform that built the .so files for Linux, so they may not cause errors. 

However, it might be a tiny bit awkward if the files were bundled for macosx but not for Windows or Linux. TBH, I don't know enough about the Mac universe to hazard a guess.

Anyway, I'm +1 on this overall.

keitherskine:
As mentioned over on issue 1082, my thoughts on this PR after all the discussion are as follows:

- include Windows 32-bit wheels, because they can be difficult to compile on your PC
- don't include unixODBC in the Mac wheels (if that's possible, I need to check)
- stop providing Linux wheels altogether, because pyodbc should ideally be built from source anyway on Linux
- also, update the GitHub Action helpers from V2 to V3

I agree we should spruce up the Wiki with better information about installing pyodbc for all OS's, but particularly for Linux and its many flavours.  We should update the README as well.  As ever, thoughts welcome.

gordthompson:
One point in favour of providing Linux wheels (without the bundled libraries, that is):

Building pyodbc from source requires build tools, and installing them can consume a non-trivial amount of disk space. Not the several gigabytes of Visual Studio goodness required to build pyodbc on Windows, but still …

Users of distros like Alpine Linux might be particularly vocal about this. On my Alpine Linux VM with

```
apk install python3
python3 -m ensurepip
```

`df -h` shows 178.9M used on `/`. After I do

```
apk add python3-dev g++ unixodbc-dev
```

the used space has increased to 393.3M. So installing just the stuff to build pyodbc more than doubles the used space.

gordthompson:
Also, re: 32-bit Windows wheels - If they're easy enough for us to provide, then it's probably a good idea. There could very well be quite a few older oddball Windows ODBC drivers that are 32-bit only, so if somebody needs to use one then at least we could save them the trouble of having to build pyodbc too.

keitherskine:
Good to hear everybody's thoughts on this.  They are greatly appreciated.  Further to that, I've updated this PR with what I hope is the general prevailing opinions:

- Include 32-bit Windows builds.
- Include ARM builds for macOS.
- Base the Linux builds on CentOS 7, not Debian 9, because it is older and hence should have wider compatibility.  CentOS installs unixODBC 2.3.1 by default but that should be good enough for our purposes.
- Don't embed any libraries in the wheels.
- Finally, I've included the standard "source distribution" for good measure, may as well generate all the distributions at the same time.

Please take a look at the artifacts generated by this latest commit and give your thoughts on them:

https://github.com/mkleehammer/pyodbc/actions/runs/2757580203

gordthompson (approved):
Both 32-bit and 64-bit Windows wheels for python 3.10 verified on Windows 8.1. Linux wheel for Python 3.6 verified on Ubuntu 18.04. 

Looks good to me.

keitherskine:
Many thanks for checking the latest builds against various platforms @gordthompson .  I have checked the macOS Intel builds and I'm just waiting for a friend to confirm the M1 builds.

keitherskine:
I haven't heard back from my friend about M1's but I'm going to merge this PR anyway.  If we need to address this further, we can do so before the next release.

PR #1089, opened 2022-08-04
Use argparse instead of optparse

keitherskine:
The standard library package `optparse` has been [deprecated](https://docs.python.org/3/library/optparse.html) since Python 3.2.  It will be removed in Python 3.12.  Hence, this PR replaces the use of `optparse` with `argparse` (but not in Python 2.7 code).

### DIFF
--- a/tests3/accesstests.py
+++ b/tests3/accesstests.py
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 
 usage="""\
-usage: %prog [options] filename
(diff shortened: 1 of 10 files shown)


gordthompson (approved):
Just tested with tests3/sqlservertests.py on Ubuntu 20.04 and it seems to be working fine.

keitherskine:
Thanks @gordthompson !  Much appreciated.

PR #1119, opened 2022-11-06
Run unit tests for Python 3.11 with Appveyor

keitherskine:
Now that Appveyor has added Python 3.11 to their Windows image, we can run the unit tests against Python 3.11 on Windows.

Also, a tweak to the CICD pyodbc install commands to not use the soon-to-be-deprecated form ("python setup.py install").


gordthompson (approved):
Looks good to me.

PR #1126, opened 2022-11-19
Separate the artifacts workflow from the build test workflow

keitherskine:
The existing Github Workflow "Ubuntu build" currently does two things on each and every commit - run the unit tests and build the artifacts for a release.  Running the unit tests is relatively quick and should happen on each commit, but the artifacts (sdist and the wheels) need to be built only at the time of a release.  Besides, building the artifacts each time is slow, wasteful, and unnecessary.  Hence, I have separated out the artifacts build into a separate workflow.  The intention is that this new workflow will be run manually when needed.  Once this PR is merged into `master`, a button will appear on the Github Actions portal that allows you to run the workflow directly.  See [here](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) for details.  I've also updated the new workflow to not call `setup.py` directly, this is [discouraged](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).  Instead, it uses the `build` module.

Finally, tweaked `README.md` and `MANIFEST.in`.  In `MANIFEST.in`, the `global-exclude` command should happen last (the commands are processed in order).


mkleehammer:
This is a great idea.

PR #1127, opened 2022-11-21
Add typing for Connection, Cursor, and connect parameters

baluyotraf:
Hello!

We are using pyright with pyodbc in our library. On strict settings, pyright complains about some of the functions/classes that they have unknown typing. Adding the types on variable arguments and in `attrs_before` fixes the issue. 

The `attrs_before` keys and values right typing looks `Dict[str, Any]` but I started the commit with `Dict[Any, Any]` to be safe.


src/pyodbc.pyi
```diff
@@ -451,5 +451,5 @@ def connect(connstring: Optional[str] = None,
             ansi: bool = False,
             readonly: bool = False,
             timeout: int = 0,
-            attrs_before: dict = {},
-            **kwargs) -> Connection: ...
+            attrs_before: Dict[Any, Any] = {},
```
keitherskine, line 454: Could we change this to `Dict[int, Any]`?  The dict keys on `attrs_before` must be integers.  The dict values can be a variety of types so best leave that as `Any` for now.

keitherskine (changes requested):
Many thanks for this @baluyotraf !  Just one small tweak needed, if you would.

src/pyodbc.pyi
baluyotraf: Done.

PR #1130, opened 2022-11-24
database-specific unit test folders, updated test scripts

keitherskine:
Note, this PR will not work with Python 2.7!  Do not merge until pyodbc no longer supports Python 2.7.

The most significant part of this PR is that the unit test scripts have been moved into database-specific folders.  There is now a separate folder for each database.  The `pytest` utility is now directed to the relevant folder to find ("discover") all the unit tests for a specific database.  Hence, in the future, new unit tests can be added to the relevant database folder simply as brand new scripts, rather than having to add the tests to the existing database-specific test script.  `pytest` will find them all.

Other updates:

- Python 2.7 and 3.6 are dropped from the CI scripts (but this PR assumes they are already no longer supported)
- the Ubuntu build tests now use the new `run_tests()` utility function, so that all tests happen from one command
- for the Windows build tests on Appveyor:
  - SQL Server 2019 is now used (instead of SQL Server 2017)
  - PostgreSQL 13 is now used (instead of PostgreSQL 11 which is rather old now), including the drivers
  - the latest MySQL 8.0 drivers are now used
  - the updated `run_tests()` utility function is used to run tests
  - the `build` package is used to generate wheels instead of `setup.py`, to be more PEP517-compliant

Notes:

- The unit test scripts have not been changed to use "pytest" style unit tests.  The scripts still use "unittest" style unit tests.  That work will be the subject of a later PR.
- When Python 2.7 is dropped, the "tests2" folder will be deleted and the "tests3" folder will be renamed to just "tests".  At that point, the testing code will have to be updated accordingly.


mkleehammer:
@keitherskine Is this still something we want?  There are conflicts and I haven't pulled it down to look into it yet.  You made a lot of other changes later, so I suspect this may no longer be needed.

keitherskine:
I converted this PR to a draft PR deliberately so it couldn't be merged.  I think it's now obsolete (in which case I will close it), but let me check if there are any good ideas in it that should still be salvaged.

PR #1134, opened 2022-12-16
Add docstrings and more to pyodbc.pyi

keitherskine:
This PR enhances the existing `pyodbc.pyi` file by:

- adding docstrings to the module functions, classes, class methods, and class properties
- adding missing parameters to various methods that didn't have them before
- making corrections to a few function signatures
- setting class properties correctly with `@property` (specifically r/o or r/w)

The docstrings are in [Google format](https://google.github.io/styleguide/pyguide.html), although I should say I haven't stuck rigidly to that guide.

The docstrings I've added don't attempt to provide a full description for every function.  The behavior of many functions is complex and I believe the Wiki is still the best place to provide comprehensive documentation.  My intention is to enhance the experience for developers by adding more summary info to the popups often seen in PyCharm, VSCode, etc.

I'm aware that `List[]` and `Dict[]` can generally be used in their base form (`list[]` and `dict[]`) but I stuck with `List[]` and `Dict[]` in this PR for consistency.  Ditto for `Optional` and `| None`.

This PR does make use of the relatively new `typing.TypeAlias` annotation which is available only in 3.10+ (also `typing.Final` from 3.8).  I don't think this should be an issue because this shouldn't affect running code and linters should probably be using the latest version of Python anyway, but if anybody thinks this is problematic, let me know.

Also, the build is failing for Python 3.6 because Github Action's runner `ubuntu-latest` was recently upgraded from 20.04 to 22.04, and 22.04 is not available for Python 3.6 in Github Actions.  (Another reason for dropping 2.7/3.6.)

### DIFF
--- a/src/pyodbc.pyi
+++ b/src/pyodbc.pyi
@@ -1,5 +1,13 @@
 from __future__ import annotations
-from typing import Any, Callable, Dict, Generator, Iterator, List, Optional, Sequence, Tuple, Union
+from typing import (
+    Any, Callable, Dict, Final, Generator, Iterable, Iterator,
+    List, Optional, Sequence, Tuple, TypeAlias, Union,
+)
+
+# ref: https://peps.python.org/pep-0249/#description
+FieldDescription: TypeAlias = Tuple[str, Any, int, int, int, int, bool]
+RowDescription: TypeAlias = Tuple[FieldDescription]
+
 
 
 # SQLSetConnectAttr attributes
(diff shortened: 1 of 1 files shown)


src/pyodbc.pyi
```diff
@@ -324,97 +334,587 @@ class ProgrammingError(DatabaseError): ...
...
-    def close(self) -> None: ...
+        Args:
+            encoding: Text encoding codec, e.g. "utf-8".
+            ctype: The C data type when passing data - either pyodbc.SQL_CHAR or pyodbc.SQL_WCHAR.  More
+                relevant for Python 2.7.
+        """
+        ...
 
+    def setdecoding(self,
+                    sqltype: int,
```
AlexandreArpin, line 402: Was the Literal type annotation considered for "enum-like" values such as this one? (See [typing.Literal](https://docs.python.org/3/library/typing.html#typing.Literal))

keitherskine: The Literal annotation requires literal integer values like 25, -123, etc. rather than variable names like "SQL_WCHAR".  The ODBC attributes like SQL_WCHAR and SQL_DRIVER_NAME are not defined by pyodbc as such, the values are part of the underlying ODBC protocol, and these values are generally not exposed.  For example, SQL_CHAR is defined in `sql.h` as value 1.  SQL_WCHAR is defined in `sqlucode.h` as value -8.  But nobody ever uses 1 and -8 in code.  Although defining these Python parameters as Literals is a noble aim, I'm not sure it's truly practical.

AlexandreArpin: Excellent reason, thanks a lot for your work on this project :)

gordthompson (approved):
PyCharm seems to be picking up the docstrings when I do 

View > Quick Definition

so that part seems to be working, at least. (Although I understand that PyCharm can be a bit … odd … in what it recognizes and what it doesn't.)

Issue #837, opened 2020-11-02
Wheel support for linux aarch64


odidev:
**Summary**
Installing pyodbc on aarch64 via pip using command "pip3 install pyodbc" tries to build wheel from source code.

**Problem description**
pyodbc doesn't have wheel for aarch64 on PyPI repository. So, while installing pyodbc via pip on aarch64, pip builds wheel for same resulting in it takes more time to install pyodbc. Making wheel available for aarch64 will benefit aarch64 users by minimizing pyodbc installation time.

**Expected Output**
Pip should be able to download pyodbc wheel from PyPI repository rather than building it from source code.

@pyodbc-team, please let me know if I can help you building wheel/uploading to PyPI repository. I am curious to make pyodbc wheel available for aarch64. It will be a great opportunity for me to work with you.

mkleehammer:
@keitherskine Is this something we can get from Travis-CI integration?  They let us test in different environments - can we get artifacts from those tests?

keitherskine:
As far as I can tell, Travis does have the ability to [upload](https://docs.travis-ci.com/user/uploading-artifacts/) artifacts to a cloud provider of your choice (e.g. AWS S3, PyPI, and many others) at the end of the CI job cycle.  Unlike AppVeyor though, Travis doesn't seem to have support for artifacts built into its web portal so you have to provide your own credentials for the cloud provider.  Travis will upload the files to wherever you want them, but do nothing more.  Also, if we want to generate artifacts for aarch64 in Travis, we'd probably have to run all the unit tests on aarch64 too.  That's not necessarily an issue, but I don't think Travis is designed to just build artifacts on specific platforms.

On the wider issue of generating builds for multiple platforms, perhaps a better approach might be to use [cibuildwheel](https://github.com/joerick/cibuildwheel) to do the heavy lifting.  ([Here's](https://github.com/scikit-learn/scikit-learn/blob/main/.travis.yml) an example from the scikit-learn project of building aarch64 on Travis with cibuildwheel.)

mkleehammer:
I forgot about cibuildwheel.  Great idea.

AlexandreArpin:
Noticed that pyodbc now uses cibuildwheel, does it mean it should be fairly easy to add linux aarch64 wheels?

I'd love to contribute but I'm fairly inept in everything Travis/AppVeyor

keitherskine:
[...]

PR #1135, opened 2022-12-16
Add aarch64 for linux

keitherskine:
Update the CI pipeline to build wheels for ARM64 Linux processors.  As requested in #837 .

Whilst we're here, remove the `push`/`pull_request` triggers for this workflow.  We don't need to build wheels every time a commit is pushed.  Keep `workflow_dispatch` so that this workflow can be manually triggered at release time.


keitherskine:
Had to make a couple of changes to get the tests to pass.

- In Github Actions, `ubuntu-latest` is now 22.04 which isn't available for Python 3.6, so pinned the runner to 20.04 for the unit tests.
- In AppVeyor, for some reason I still don't understand, downloading MySQL drivers no longer works with the standard PowerShell command `Invoke-WebRequest` (getting 403's).  Downloading drivers for SQL Server and Postgres drivers is still OK, just not MySQL.  Weird.  Using `curl` seems to work so implemented that instead.  `curl` also has the advantage of having the `--verbose` option which `Invoke-WebRequest` doesn't have.

As well as these two fixes, I've added various other tweaks to make the AppVeyor log output better, and to make sure the AppVeyor cache of driver msi files for the jobs is kept up-to-date and correct.

This PR should be good to review now.

appveyor/install.ps1
```diff
@@ -1,24 +1,33 @@
 # check that all the required ODBC drivers are available, and install any that are missing
 
 Function DownloadFileFromUrl ($url, $file_path) {
+    $curl_params = "-f -sS -L -o `"$file_path`" `"$url`""
+    If (${env:APVYR_VERBOSE} -eq "true") {
```
AlexandreArpin, line 5: nitpick:  all the other code uses the lowercase form of built-in functions.

```suggestion
    if (${env:APVYR_VERBOSE} -eq "true") {
```

gordthompson (approved):
lgtm

appveyor/install.ps1
gordthompson: +1 - may as well be consistent

keitherskine: Will do.

PR #1148, opened 2023-01-09
Add typing definition for Row to be a valid Iterable

baluyotraf:
Using `Row` as an iterable in code results in this typing issue. For example:

```python
zip(row, row)
```

The code above works fine, but it gives the typing error below

```
Argument of type "Row" cannot be assigned to parameter "__iter2" of type "Iterable[_T2@__new__]" in function "__new__"
  "Row" is incompatible with protocol "Iterable[_T2@__new__]"
    "__iter__" is not present
```

This can be worked around currently by casting the `Row` object manually

```python
from typing import cast, Iterable, Any

irow = cast(Iterable[Any], row)
```

This fix simply adds the `__iter__` method in the typing definition.


keitherskine:
Many thanks for this @baluyotraf .  I'm kinda surprised `iter()` on a Row object runs at all.  After all, the underlying function "tp_iter" hasn't actually been coded up.  But hey, it works!  So yes, adding `__iter__` to the Row class to formalize this is a good idea.

Issue #1186, opened 2023-03-20
`typing.TypeAlias` causing type-review issues in Ubuntu 20.04 LTS


ajnelson-nist:
Hello,

[PR 1134](https://github.com/mkleehammer/pyodbc/pull/1134) included this remark:

> This PR does make use of the relatively new `typing.TypeAlias` annotation which is available only in 3.10+ (also `typing.Final` from 3.8). I don't think this should be an issue because this shouldn't affect running code and linters should probably be using the latest version of Python anyway, but if anybody thinks this is problematic, let me know.

I'm unfortunately late in arriving to this change, and found that it does entail a slight operational complexity.

The Microsoft "Quick start" documentation for SQL Server on Ubuntu still uses Ubuntu 20.04.  (Documentation is [here](https://learn.microsoft.com/en-us/sql/linux/quickstart-install-connect-ubuntu?view=sql-server-ver16)).  That edition of Ubuntu defaults to Python 3.8, has 3.9 available, and doesn't officially provide 3.10 or 3.11 (per [Ubuntu package search results](https://packages.ubuntu.com/search?keywords=python3.10), as well as `apt` being unable to locate `python3.10`).  So, the population of users that follow SQL Server on Ubuntu documentation are adversely effected by `TypeAlias` from PR 1134.

Further, Python 3.8 has a scheduled end of support in October of 2024 (per, among other sources, the [Python.org downloads page](https://www.python.org/downloads/)).

I happen to be working on an application that needs to do some steps in Python 3.8, using SQL Server in environments including Ubuntu 20.04.  Part of its testing workflow is relying on type checkers to ensure, for example, that I didn't introduce a feature in my own programming only available in a newer version of Python.

Is it possible to revert the changes in PR 1134 that are not compatible with Python 3.8?  The key incompatibility I'm hitting is this, reported by `mypy` (not `mypy --strict`):

```
src/pyodbc.pyi:2: error: Module "typing" has no attribute "TypeAlias"  [attr-defined]
```

[...]

PR #1191, opened 2023-04-02
Don't use py3.10 TypeAlias in pyodbc.pyi

keitherskine:
Re #1186 , the typing construct `TypeAlias` was introduced in Python 3.10.  This is causing mypy issues with people who are using older versions of Python.  Hence, drop the `TypeAlias` references in `pyodbc.pyi` and just use the raw typing definition for cursor descriptions.  This appears to solve the mypy issue:

```
(pyodbcpyi38) C:\Users\toast>mypy C:\Users\toast\Documents\GitHub\forks\pyodbc\src\pyodbc.pyi
Success: no issues found in 1 source file
```

Note, the `Final` typing references are being kept, despite them being introduced only in Python 3.8.

### DIFF
--- a/src/pyodbc.pyi
+++ b/src/pyodbc.pyi
@@ -1,13 +1,8 @@
 from __future__ import annotations
 from typing import (
     Any, Callable, Dict, Final, Generator, Iterable, Iterator,
-    List, Optional, Sequence, Tuple, TypeAlias, Union,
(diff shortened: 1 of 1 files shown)


ajnelson-nist (approved):
I've tested this, and confirm this addresses the issues I raised in #1186.  Thank you!

gordthompson (approved):
lgtm

PR #1207, opened 2023-04-21
Allow python exception handling for empty iterator

ndmlny-qs:
This commit allows Python to handle empty iterators given to a `pyodbc` connection instead of issuing a segfault. See the below code for an example of what would normally result in a segfault.

```python
import collections
import pyodbc

class MySequence(collections.abc.Sequence):
    def __getitem__(self, index):
        ...

    def __len__(self):
        return 1

connection = pyodbc.connect(...)
connection.execute("SELECT ?, ?", 123, MySequence()).fetchone()
```

### DIFF
--- a/src/params.cpp
+++ b/src/params.cpp
@@ -1177,11 +1177,6 @@ static bool GetByteArrayInfo(Cursor* cur, Py_ssize_t index, PyObject* param, Par
 // TVP
 static bool GetTableInfo(Cursor *cur, Py_ssize_t index, PyObject* param, ParamInfo& info)
 {
-    if (!PyIter_Check(param))
-    {
-        PyErr_SetString(PyExc_ValueError, "The given sequence was empty.");
-        return false;
-    }
     int nskip = 0;
     Py_ssize_t nrows = PySequence_Size(param);
     if (nrows > 0)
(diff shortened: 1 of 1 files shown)


mkleehammer:
Well, we certainly don't want any segfaults.  Appveyor builds failed for some reason.  It looks like the new exception is being raised in all of the SQL Server tests or something.

Once that is worked out, I'm wondering if `ValueError` would be more appropriate.

ndmlny-qs:
I see where the tests need to be updated, and will have to look at them later next week for a fix.

The tests do show a `ValueError` being raised, https://ci.appveyor.com/project/mkleehammer/pyodbc/builds/46853185/job/tkldqoprtygxr1vp#L702, unless I'm reading that incorrectly. I'm new to the Python ABI so if I should have chosen a different way to raise the exception, then I'm happy to switch it. Just let me know what you would like changed and I'll fix it.

ndmlny-qs:
I modified the segfault so that the return propagates back to the Python code that calls it. We have GitHub mimic Appveyor tests here https://github.com/Quansight/pyodbc/actions/runs/5592899278 and they all were green, except for Python 2.7, as it is deprecated. I'm not sure why the tests here do not pass. If you have any ideas, I'm happy to spend time investigating and implementing any fixes you would like done.

ndmlny-qs:
closing so I can use the 5.0 version of pyodbc for this update

PR #1212, opened 2023-04-27
Use `HOMEBREW_PREFIX` to locate Homebrew

### LINKED ISSUE #1211: setup.py doesn't use `HOMEBREW_PREFIX` to find Homebrew (sometimes it can't find it)
### Issue

Homebrew installs in either a default directory (based on factors such as the macOS version installing Homebrew, the Homebrew version being installed, etc.) or a user-specified directory.

Homebrew provides a `HOMEBREW_PREFIX` environment variable in order to locate where it's installed.

While setup.py takes a few guesses as to where Homebrew is located ([setup.py#L213-L219]), it does not use the environment variable which is likely to have the location:

https://github.com/mkleehammer/pyodbc/blob/7b4e27781c7f1b75995ecca73d3018c929f4adef/setup.py#L213-L219

[setup.py#L213-L219]:
https://github.com/mkleehammer/pyodbc/blob/7b4e27781c7f1b75995ecca73d3018c929f4adef/setup.py#L213-L219

If the user installed Homebrew in a user-specified directory, setup.py is not able to find Homebrew and fails to run.

RA80533:
Fixes #1211

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -217,6 +217,9 @@ def get_compiler_settings(version_str):
                 '/opt/homebrew/include',
                 expanduser('~/homebrew/include'),
             ]
+            _HOMEBREW_PREFIX = os.environ.get('HOMEBREW_PREFIX')
+            if _HOMEBREW_PREFIX is not None:
+                dirs.insert(0, os.path.join(_HOMEBREW_PREFIX, 'include'))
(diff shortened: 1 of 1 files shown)


RA80533:
@mkleehammer What can I do to bring this over the finish line?

PR #1214, opened 2023-04-27
Try-catch `os.popen()` for command failures

### LINKED ISSUE #1213: setup.py fails if `odbc_config` can't be found
### Issue

setup.py attempts to run `odbc_config` in the user's shell via `os.popen()`:

https://github.com/mkleehammer/pyodbc/blob/7b4e27781c7f1b75995ecca73d3018c929f4adef/setup.py#L204

It then calls `readlines()` and assumes there will be at least 2 values to unpack and assign to `cflags` and `ldflags`:

https://github.com/mkleehammer/pyodbc/blob/7b4e27781c7f1b75995ecca73d3018c929f4adef/setup.py#L205

If `odbc_config` isn't found in the user's PATH, `readlines()` won't unpack anything and a ValueError will be raised. This issue is likely often to occur when a user is attempting to install pyodbc where it unfortunately produces an unhelpful error:

```console
(example) RA80533@mbp example % python -m pip install --pre --no-binary :all: pyodbc
Collecting pyodbc
  Using cached pyodbc-4.0.39.tar.gz (282 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [20 lines of output]
        File "/Users/RA80533/Desktop/example/.venv/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
          main()
        File "/Users/RA80533/Desktop/example/.venv/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/Users/RA80533/Desktop/example/.venv/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^
        File "/private/var/folders/rx/mgpqyfxd01q4w5yhxdctdfcm0000gn/T/pip-build-env-b_hshvm7/overlay/lib/python3.11/site-packages/setuptools/build_meta.py", line 341, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=['wheel'])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/private/var/folders/rx/mgpqyfxd01q4w5yhxdctdfcm0000gn/T/pip-build-env-b_hshvm7/overlay/lib/python3.11/site-packages/setuptools/build_meta.py", line 323, in _get_build_requires
          self.run_setup()
        File "/private/var/folders/rx/mgpqyfxd01q4w5yhxdctdfcm0000gn/T/pip-build-env-b_hshvm7/overlay/lib/python3.11/site-packages/setuptools/build_meta.py", line 338, in run_setup
          exec(code, locals())
        File "<string>", line 371, in <module>
        File "<string>", line 77, in main
        File "<string>", line 205, in get_compiler_settings
      ValueError: not enough values to unpack (expected 2, got 0)
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
```

RA80533:
Fixes #1213

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -202,8 +202,12 @@ def get_compiler_settings(version_str):
 
         # Homebrew installs odbc_config
         pipe = os.popen('odbc_config --cflags --libs 2>/dev/null')
-        cflags, ldflags = pipe.readlines()
-        exit_status = pipe.close()
+        try:
+            cflags, ldflags = pipe.readlines()
+        except ValueError:  # The command failed
+            cflags, ldflags = '', ''
+        finally:
+            exit_status = pipe.close()
 
         if exit_status is None:
             settings['extra_compile_args'].extend(shlex.split(cflags))


setup.py
```diff
@@ -123,17 +194,20 @@ def get_compiler_settings():
...
         ])
 
         # Homebrew installs odbc_config
         pipe = os.popen('odbc_config --cflags --libs 2>/dev/null')
-        cflags, ldflags = pipe.readlines()
-        exit_status = pipe.close()
+        try:
+            cflags, ldflags = pipe.readlines()
+        except ValueError:  # The command failed
+            cflags, ldflags = '', ''
```
bkline, line 208: I assume the patch is setting `cflags` and `ldflags` to empty strings in the `except` block because you want to guard against another exception in the case in which we end up in the `exit_status is None` block. Presumably that would happen if `odbc_config` was found but produced malformed output for some reason. In that case wouldn't it be preferable to at least warn the user about the problem, if not abort the script altogether with an appropriate error message? You could set `cflags` and `ldflags` to `None` and then test them in the `if `exit_status is None` block.

PR #1245, opened 2023-07-25
Get the diagnostics messages more often

### LINKED ISSUE #1238: PRINT messages from triggers not being returned when SET NOCOUNT is ON
Please first make sure you have looked at:

* Documentation: https://github.com/mkleehammer/pyodbc/wiki
* Other issues

### Environment

- Python: 3.11
- pyodbc: 4.0.39
- OS: Windows
- DB: SQL Server 2019
- driver: ODBC Driver 18 for SQL Server 2018.181.02.01

### Issue

The cursor.messages is always empty in the below code because of the **SET NOCOUNT ON** statemen but I see the PRINT statement when I run it in Microsoft SQL Studio Manager. I believe it's because you're not calling [SQLGetDiagRec](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlgetdiagrec-function?view=sql-server-ver16) when you get **SQL_NO_DATA_FOUND**.

According to [Return Codes ODBC](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/return-codes-odbc?view=sql-server-ver16). The application should call **SQLGetDiagRec** or **SQLGetDiagField** to retrieve additional information when it gets the following return codes.

- SQL_SUCCESS
- SQL_SUCCESS_WITH_INFO
- SQL_ERROR
- SQL_NO_DATA
- SQL_NEED_DATA
- SQL_STILL_EXECUTING

So please make sure you're calling those functions in those scenarios.

### Code to replicate

``` python
import logging
import unittest

import pyodbc

log = logging.getLogger(__name__)


class MyTestCase(unittest.TestCase):

    def test_something(self):
        conn_str = "Driver=ODBC Driver 18 for SQL Server;Server=localhost;TrustServerCertificate=yes;Trusted_Connection=yes;APP=tests.utils.TestCase:1022::46840;MARS_Connection=yes;ConnectRetryCount=30"
        conn = pyodbc.connect(conn_str, autocommit=True)
        cursor = conn.execute("DROP TABLE IF EXISTS FOO; CREATE TABLE FOO(ID INT IDENTITY PRIMARY KEY, NAME NVARCHAR(255) NULL);")
        while cursor.nextset():
            cursor.nextset()
        cursor = conn.execute("""
CREATE  TRIGGER [dbo].[trg_FOO] ON FOO
AFTER INSERT
AS
BEGIN
    SET NOCOUNT ON
    --UNCOMMENTING THE BELOW LINE CAUSES PYODBC TO GET THE PRINT STATEMENT
    --SET NOCOUNT OFF
    PRINT 'WOW!!!!'
END
        """)
        cursor.nextset()
        cursor = cursor.execute("INSERT INTO FOO(Name) SELECT NEWID() WHERE 1=0")
        log.info(f"{cursor.messages=}")
        while cursor.nextset():
            log.info(f"{cursor.messages=}")
        log.info(f"{cursor.messages=}")


if __name__ == '__main__':
    unittest.main()
```

### Trace Files
[SQL.NO_COUNT_OFF.LOG](https://github.com/mkleehammer/pyodbc/files/12004007/SQL.NO_COUNT_OFF.LOG)
[SQL.NO_COUNT_ON.LOG](https://github.com/mkleehammer/pyodbc/files/12004008/SQL.NO_COUNT_ON.LOG)

keitherskine:
Diagnostic messages (e.g. SQL error messages) are retrieved after executing a SQL query.  However, the diagnostic messages were being retrieved only for a return value of SQL_SUCCESS_WITH_INFO.  This should be extended for other return values.

Function GetDiagRecs has been changed to add diagnostic messages to the "messages" cursor attribute rather than replace the attribute.  Deciding when to clear the "messages" attribute should happen outside GetDiagRecs.

Fixes #1238 .

Note, Github Actions [no longer supports Python 2.7](https://github.com/actions/setup-python/issues/672).

Ref:
https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/return-codes-odbc
https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlexecdirect-function?view=sql-server-ver16#returns
https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlmoreresults-function?view=sql-server-ver16#returns

### DIFF
--- a/.github/workflows/ubuntu_build.yml
+++ b/.github/workflows/ubuntu_build.yml
@@ -11,8 +11,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - python-version: "2.7"
-            tests-dir: tests2
           - python-version: "3.6"
             tests-dir: tests3
           - python-version: "3.7"
@@ -161,7 +159,7 @@ jobs:
     - uses: actions/checkout@v3
 
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
 
--- a/src/cursor.cpp
+++ b/src/cursor.cpp
@@ -580,9 +580,7 @@ static bool PrepareResults(Cursor* cur, int cCols)
 
 static int GetDiagRecs(Cursor* cur)
 {
-    // Retrieves all diagnostic records from the cursor and assigns them to the "messages" attribute.
-
-    PyObject* msg_list;  // the "messages" as a Python list of diagnostic records
+    // Retrieves all diagnostic records from the cursor and adds them to the "messages" attribute.
 
     SQLSMALLINT iRecNumber = 1;  // the index of the diagnostic records (1-based)
     ODBCCHAR    cSQLState[6];  // five-character SQLSTATE code (plus terminating NULL)
@@ -599,9 +597,12 @@ static int GetDiagRecs(Cursor* cur)
       return 0;
     }
 
-    msg_list = PyList_New(0);
-    if (!msg_list)
-        return 0;
+    // just in case the "messages" attributes isn't a list (i.e. empty or null)
+    if (!cur->messages || !PyList_Check(cur->messages))
+    {
+        Py_XDECREF(cur->messages);
+        cur->messages = PyList_New(0);
+    }
 
     for (;;)
     {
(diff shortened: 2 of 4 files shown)


gordthompson (commented):
I have nothing further to add to this issue.

mkleehammer:
This is being referenced in #1238.  I really don't remember this, but it is closed.  Is this something that I need to reimplement?  @keitherskine

PR #1256, opened 2023-08-27
Remove old Python27 stuff from pyodbc.pyi

keitherskine:
Looks like a couple of entities are no longer in the pyodbc interface - "ansi" and "UNICODE_SIZE".  This PR removes them.

### DIFF
--- a/src/pyodbc.pyi
+++ b/src/pyodbc.pyi
@@ -296,7 +296,6 @@ SQL_XOPEN_CLI_YEAR: int
 
 # pyodbc-specific constants
 BinaryNull: Any  # to distinguish binary NULL values from char NULL values
-UNICODE_SIZE: int
 SQLWCHAR_SIZE: int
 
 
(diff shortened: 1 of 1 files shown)


mkleehammer:
And again, thanks for noticing stuff I didn't.  Is there any way to create a unit test to compare the pyi file against the binary?

Long term I'm thinking we might want to split pyodbc into Python classes and a _pyodbc C library, as long as it doesn't degrade performance.  A performance benchmark is necessary for determining if the stable ABI is a good idea too.  Hmmm...  Need to prioritize that.

keitherskine:
The utility `mypy` has a facility for doing a certain amount of stub checking, but I suspect that might depend on a pure-Python underlying module.

I think creating a pure-Python top/level module for pyodbc and C underlying extension classes would be a great idea.  There's a lot of fairly straightforward non-ODBC stuff that could be done without invoking the power of C, e.g. parameter validation.  It would be handy to be able to do that in Python.

PR #1257, opened 2023-08-27
Update ODBC driver versions on AppVeyor

keitherskine:
Whilst we're jazzing things up, I thought we could update the ODBC drivers used in AppVeyor.

Note, the MySQL 5.8 driver was only needed for Python 2.7 so it can be removed.


mkleehammer:
You are on fire today.  I forgot to look in that file.  Well, I glanced at it, but not being a PS person nothing jumped out at me.  I should have thought of looking for Python 2 stuff there.

PR #1265, opened 2023-08-29
Run the CICD tests

keitherskine:
The CICD pipelines were running only the PostgreSQL and MySQL unit tests on Linux, because the tests were not being run using pytest.  This PR fixes that.  Also, strictly speaking, pytest should be run as a module with the `-m` option (i.e. `python -m pytest`) so I made those amendments too.

There is one test that is failing and I don't know why.  Here is the output:
```
__________________________ test_geometry_null_insert __________________________
cursor = <pyodbc.Cursor object at 0x0364EDA0>
    @pytest.mark.skipif(sys.platform.startswith('linux'),
                        reason='SQL Server Linux does not support -151 yet')
    def test_geometry_null_insert(cursor: pyodbc.Cursor):
        cnxn = connect()
    
        def convert(value):
            return value
    
        cnxn.add_output_converter(-151, convert)  # -151 is SQL Server's geometry
        cursor.execute("create table t1(n int, v geometry)")
        cursor.execute("insert into t1 values (?, ?)", 1, None)
>       value = cursor.execute("select v from t1").fetchone()[0]
E       pyodbc.ProgrammingError: ('ODBC SQL type -151 is not yet supported.  column-index=0  type=-151', 'HY106')
tests\sqlserver_test.py:1273: ProgrammingError
```
On the face of it, the code looks correct.  The only thing I can think of that -151 is a negative number and hence might throw things off somewhere, but I'm just guessing.  It's always a SQL_SMALLINT so would appear to be fine.  Help would be appreciated debugging this.

### DIFF
--- a/.github/workflows/ubuntu_build.yml
+++ b/.github/workflows/ubuntu_build.yml
@@ -190,32 +190,32 @@ jobs:
         PYODBC_POSTGRESQL: "DRIVER={PostgreSQL Unicode};SERVER=localhost;PORT=5432;UID=postgres_user;PWD=postgres_pwd;DATABASE=test"
       run: |
         cd "$GITHUB_WORKSPACE"
-        pytest "./tests/postgresql_test.py"
+        python -m pytest "./tests/postgresql_test.py"
 
     - name: Run MySQL tests
       env:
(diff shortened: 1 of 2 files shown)


mkleehammer:
The negative number is OK - SQL Server uses them for vendor specific extensions which we track in dbspecific.h

It turns out the test isn't really needed anymore so I've deleted it.

I rewrote the internals of managing output converters to use a Python dictionary.

PR #1266, opened 2023-08-29
Location of pyodbc.pyi on install

keitherskine:
The stub file pyodbc.pyi needs special care when generating the wheel, to ensure it is placed in the "site-packages" directory on install.  The relevant setup.py code was reverted recently, so I have re-reverted it back.  Apparently, we also need to set "include package data" to False now (because it is defaulted to True when using pyproject.toml).

FYI, this code (or at least its equivalent) does not work in pyproject.toml, hence it must remain in setup.py for the time being.

Also, quick fix for MANIFEST.in.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,7 +1,6 @@
 include src/*.h
 include src/*.cpp
-include tests2/*
-include tests3/*
+include tests/*
 include README.*
 include LICENSE.txt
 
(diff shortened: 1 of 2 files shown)


mkleehammer:
Wow Python packaging is a mess (still!).  This is why people like cargo :)

Thanks for catching this.

PR #1269, opened 2023-08-30
Suggest to install unixodbc instead of build fails with TypeError expected 2 arguments got 0. 

Carreau:
It's not much, but I guess it could prevent a few minutes of googling.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,7 @@ import sys, os, shlex, re
 from os.path import exists, join, isdir, relpath, expanduser
 from pathlib import Path
 from inspect import cleandoc
+from shutil import which
 
 from setuptools import setup
 from setuptools.extension import Extension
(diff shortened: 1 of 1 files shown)


Carreau:
See Error in https://github.com/mkleehammer/pyodbc/discussions/1263#discussioncomment-6865709,

keitherskine:
It would certainly be good if pyodbc checked for a driver manager, but we should probably check for iODBC as well, and I'm not sure about raising an exception rather than just outputting a warning.

mkleehammer:
Two things:

Suggesting conda could really confuse things.  I would use apt or nala on Ubuntu, RH uses something else, etc.

Second, do the binary wheels work with iODBC?  Do they embed the unixodbc lib or something?  I know we had to disable that until we got ARM runners or it would embed Intel libraries in the ARM binary wheels.

Issue #1277, opened 2023-09-21
pyodbc uses odbc 3.x features instead of 3.8


jabbera:
When connection pooling is on in unixodbc there is a connection reset procedure that needs to be run: https://github.com/lurcher/unixODBC/blob/9b31cd228d01e768a7ad9126d5da8860f1ed72a9/DriverManager/SQLConnect.c#L3904 Not running this causes connections in the pool to be left tainted will all sorts of badness such as the following not being done:

```
It resets all error states and numbers (like @@error)
It stops all EC's (execution contexts) that are child threads of a parent EC executing a parallel query
It will wait for any outstanding I/O operations that is outstanding
It will free any held buffers on the server by the connection
It will unlock any buffer resources that are used by the connection
It will release all memory allocated owned by the connection
It will clear any work or temporary tables that are created by the connection
It will kill all global cursors owned by the connection
It will close any open SQL-XML handles that are open
It will delete any open SQL-XML related work tables
It will close all system tables
It will close all user tables
It will drop all temporary objects
It will abort open transactions
It will defect from a distributed transaction when enlisted
It will decrement the reference count for users in current database; which release shared database lock
It will free acquired locks
It will releases any handles that may have been acquired
It will reset all SET options to the default values
It will reset the @@rowcount value
It will reset the @@identity value
It will reset any session level trace options using dbcc traceon()
```

This feature is only enabled if the environment is set to 3.8 or higher.  In pyodbc it's hard coded to 3.X which doesn't include this vital behavior. https://github.com/mkleehammer/pyodbc/blob/ee0ad21e7ab00c4ab974f332ab0ce22bb7ad776d/src/pyodbcmodule.cpp#L299

I'm happy to author a PR to enable this if it's something you would be open to.

v-chojas:
Not all drivers support ODBC 3.8, but it does seem like a good idea to make this configurable.

jabbera:
I think each connection can have its own environment, which means its own odbc version. That seems like a major change versus allowing the default to be optionally 3.8.

Issue #48, opened 2015-05-14
with pyodbc.connect(cs_sql) as conn does not close connection after leaving the block


zyzhu2000:
with pyodbc.connect(cs_sql) as conn:
   pass
print conn

You will see that the connection is not closed.

mkleehammer:
It is not supposed to close the connection - it *commits* the connection at the end if no error was raised.  If an error was raised it will roll back.

However, there is zero reason to use a context manager in CPython once you understand how it works.  It uses reference counted garbage collection so:

```python
def func():
    cursor = connect(...).cursor()
    cursor.execute(...) // 1
    cursor.execute(...) // 2
    cursor.commit()     // 3
```

This is is just as good.  All Python objects are deleted when they go out of scope if there are no other references to them.  In this case i didn't keep a copy of the Connection around so the only reference is the Cursor.  When the function exists, either by completing or via an exception, the cursor will be deleted.  Since the cursor is the only reference to the connection, it will be deleted also.  Immediately - not "someday" like Java or Javascript.  When a connection is deleted, the connection is closed an any outstanding changes are rolled back.  (All DBs roll back uncommitted transactions but pyodbc will make a rollback call anyway.)  If the code made it past the `commit` (3), your changes are saved.  If (2) were to raise an exception, the changes from (1) will be rolled back immediately as the exception unwinds the stack.

AtomBaf:
HI mkleehammer

I disagree, the connection object should be closed after the with block. I think most of developers will think that way and the current pyodbc behaviour, they will let connections objects unclosed, thus leaving way to memory leaks.

You're right about the fact that the object memory will be cleanup after exit of a function, but if you're using the connection in another context (like a Jupyter notebook), all connection objects will stay forever. The only way to use pyodbc in this context is to use the try/except blocks, which is not very clean and don't use the [PEP343](https://www.python.org/dev/peps/pep-0343/) feature.

Thank you

keitherskine:
Also relevant: #43
It appears most python odbc modules treat context managers on connections as a way of managing commits rather than the connection itself.  I was surprised at this too.

PR #1278, opened 2023-09-21
Allow customization of default odbc version

jabbera:
Fixes https://github.com/mkleehammer/pyodbc/issues/1277

My cpp is VERY rusty. Let me know if there is a better solution for this.

### DIFF
--- a/src/pyodbcmodule.cpp
+++ b/src/pyodbcmodule.cpp
@@ -67,7 +67,13 @@ static char module_doc[] =
     "\n"
     "paramstyle\n"
     "  The string constant 'qmark' to indicate parameters are identified using\n"
-    "  question marks.";
+    "  question marks."
+    "\n"
+    "odbcversion\n"
+    "  The ODBC version number as a string, such as '3.X' for ODBC 3.X compatibility.\n"
+    "  This is a global (HENV) setting, so it can only be modified before the first\n"
+    "  connection is made. Use 3.8 if you are using unixodbc connection pooling and your\n"
+    "  drivers are all 3.8 compatible. The default is 3.X\n";
 
 PyObject* Error;
 PyObject* Warning;
@@ -296,7 +302,17 @@ static bool AllocateEnv()
         return false;
     }
 
-    if (!SQL_SUCCEEDED(SQLSetEnvAttr(henv, SQL_ATTR_ODBC_VERSION, (SQLPOINTER)SQL_OV_ODBC3, sizeof(int))))
+    SQLPOINTER defaultVersion = (SQLPOINTER)SQL_OV_ODBC3;
+    PyObject* odbcversion = PyObject_GetAttrString(pModule, "odbcversion");
+    if (PyObject_TypeCheck(odbcversion, &PyUnicode_Type)) {
+        if (PyUnicode_CompareWithASCIIString(odbcversion, "3.8") == 0)
+        {
+            defaultVersion = (SQLPOINTER)SQL_OV_ODBC3_80;
+        }
+    }
+    Py_DECREF(odbcversion);
+    
+    if (!SQL_SUCCEEDED(SQLSetEnvAttr(henv, SQL_ATTR_ODBC_VERSION, defaultVersion, sizeof(int))))
     {
         PyErr_SetString(PyExc_RuntimeError, "Unable to set SQL_ATTR_ODBC_VERSION attribute.");
         return false;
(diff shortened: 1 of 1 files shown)


jabbera:
Example code that fails with 
```
  File "/mnt/z/research_hub/work/test.py", line 21, in <module>
    connection.execute(table)
pyodbc.ProgrammingError: ('42S01', "[42S01] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]There is already an object named '#TemporaryTable' in the database. (2714) (SQLExecDirectW)")
```

without this PR and passes with it and 3.8 set.

```
import pyodbc
# pyodbc.odbcversion = "3.8" # Exception above is generated without this patch

table = """
CREATE TABLE #TemporaryTable
(
    Col1 int,
    Col2 varchar(10)
);
"""

with pyodbc.connect(connection_string) as connection:
    connection.execute(table)

with pyodbc.connect(connection_string) as connection:
    connection.execute(table)

with pyodbc.connect(connection_string) as connection:
    connection.execute(table)

with pyodbc.connect(connection_string) as connection:
    connection.execute(table)

with pyodbc.connect(connection_string) as connection:
    connection.execute(table)
```

gordthompson:
What I notice about the repro code is that it seems to assume that pyodbc explicitly closes a connection created by a context manager. That is not the case, ref: https://github.com/mkleehammer/pyodbc/issues/48

While I can reproduce the error using the original repro code (with pyodbc 4.0.39), changing it to

```python
table = """
CREATE TABLE #TemporaryTable
(
    Col1 int,
    Col2 varchar(10)
);
"""

connection = pyodbc.connect(connection_string)
connection.execute(table)
connection.close()

connection = pyodbc.connect(connection_string)
connection.execute(table)
connection.close()

connection = pyodbc.connect(connection_string)
connection.execute(table)
connection.close()

connection = pyodbc.connect(connection_string)
connection.execute(table)
connection.close()

connection = pyodbc.connect(connection_string)
connection.execute(table)
connection.close()
```

does not reproduce the error. The difference is that the original repro code results in the `SQLEndTran` call being immediately followed by a `SQLAllocHandle` call for the "new" connection

```
[ODBC][8739][1696267076.950987][SQLEndTran.c][566]
        Exit:[SQL_SUCCESS]
[ODBC][8739][1696267076.951042][SQLAllocHandle.c][395]
        Entry:
            Handle Type = 2
            Input Handle = 0x141da00
        UNICODE Using encoding ASCII 'UTF-8' and UNICODE 'UTF16LE'
```

while explicitly closing the connection adds `SQLDisconnect` and `SQLFreeHandle` calls prior to the new `SQLAllocHandle`

```
[ODBC][9042][1696267851.430281][SQLEndTran.c][566]
        Exit:[SQL_SUCCESS]
[ODBC][9042][1696267851.430313][SQLDisconnect.c][208]
        Entry:
            Connection = 0x10a9d00
[ODBC][9042][1696267851.430344][SQLDisconnect.c][302]
        Exit:[SQL_SUCCESS]
[ODBC][9042][1696267851.430374][SQLFreeHandle.c][290]
        Entry:
            Handle Type = 2
            Input Handle = 0x10a9d00
[ODBC][9042][1696267851.430408][SQLFreeHandle.c][339]
        Exit:[SQL_SUCCESS]
[ODBC][9042][1696267851.430458][SQLAllocHandle.c][395]
        Entry:
            Handle Type = 2
            Input Handle = 0x114c0d0
        UNICODE Using encoding ASCII 'UTF-8' and UNICODE 'UTF16LE'
```

I also notice that the original repro code fails on the third context manager, not the second, which makes me suspect that "closing" a connection by clobbering the existing `connection` variable with another `.connect()` call might not behave as one might assume.

TL;DR - I'm just a bit confused. Perhaps @v-chojas might have some insights on this.

jabbera:
I'll certainly wait on the expert to comment but I'll add my 2c here also. Even if the particular issue demoed can be resolved by using the connection correctly I'd think we would want the ability for the connection to be reset when it gets added back into the pool which only happens with 3.8. Some of the items it does are not covered by the endtran call. Additionally if a connection leaks (not calling close) accidentally you won't end up with a dirty connection. 

Also, while I demoed this issue using pyodbc directly I stumbled across it using sqlalchemy with the nullpool because we prefer unixodbc pooling. I would be shocked if they were missing the close call, but I'll dig into it.

jabbera:
@gordthompson now I'm really confused. With 4.0.39 this produces the error even though the connection is being closed:

```
with pyodbc.connect(connection_string) as connection:
    connection.execute(table)
connection.close()

with pyodbc.connect(connection_string) as connection:
    connection.execute(table)
connection.close()

with pyodbc.connect(connection_string) as connection:
    connection.execute(table)
connection.close()

with pyodbc.connect(connection_string) as connection:
    connection.execute(table)
connection.close()

with pyodbc.connect(connection_string) as connection:
    connection.execute(table)
connection.close()

```

jabbera:
@gordthompson I think you have a bug in your test case. Because autocommit is not enabled close in your case rolls back the transaction that creates the temporary table. If you turn on autocommit or add a commit statement the bug reproduces fine:

```
for i in range(20):
    connection = pyodbc.connect(connection_string)
    connection.execute(table)
    connection.commit()
    connection.close()
```

gordthompson (commented):
Looks good to me.

mkleehammer:
Thanks.  In the future, we might want to add an error message if the user tries something unsupported like "3.9".

Issue #1289, opened 2023-10-10
pyodbc 5.0.0 core dump on connecting with token to mssql


karoldob:
### Environment

- Python: 3.10.12
- pyodbc: 5.0.0
- sqlalchemy: 2.0.21
- OS: Ubuntu 22.04 64bit
- DB: MSSQL (AzureSQL)
- driver:  Microsoft ODBC Driver 18 for SQL Server (/opt/microsoft/msodbcsql18/lib64/libmsodbcsql-18.3.so.2.1)
- login method: token

### Issue
Trying to connect with token as described in:
https://docs.sqlalchemy.org/en/20/dialects/mssql.html#connecting-to-databases-with-access-tokens

How to reproduce:
- login with `az login` on your machine (in MS way)
- copy-paste sample code from the link and fill server/database
- add `engine.connect()` at the end
- observe segmentation fault (core dumped)

Observations:
pyodbc 4.0.38 is not affected - above samples works well
Just `pip install --upgrade pyodbc==4.0.38` solves the issue, in the same environment
Login with user-password works well

Trace in attachment
[odbctrace.log](https://github.com/mkleehammer/pyodbc/files/12862265/odbctrace.log)

karoldob:
Yes, already tested with this samples:
https://github.com/mkleehammer/pyodbc/wiki/Connecting-to-SQL-Server-from-Linux#connecting-to-servers-with-access-tokens

Behavior is the same: segfault on pyodbc 5.0.0, normal working on 4.0.38

keitherskine:
Hi @karoldob , as an experiment, could I ask you to try using a bytearray rather than bytes for the access token in your call to pyodbc.connect.  That is, instead of:
```
cnxn = pyodbc.connect(connection_string, attrs_before={SQL_COPT_SS_ACCESS_TOKEN: token_struct})
```
try:
```
cnxn = pyodbc.connect(connection_string, attrs_before={SQL_COPT_SS_ACCESS_TOKEN: bytearray(token_struct)})
```
It might work.  (But of course bytes should work too.)  Many thanks.

jacovg91:
[...]

PR #1290, opened 2023-10-11
Handle bytes in the attrs_before parameter for pyodbc.connect()

keitherskine:
An attempt to fix the segfault issue #1289 .

It looks like the code to handle byte values in the attrs_before parameter was lost in the v5 upgrade.  I've copied the original code back in.  The only difference is I've replaced the call to PyBytes_AS_STRING with the stable ABI compliant call to PyBytes_AsString.

@mkleehammer / @v-chojas / @gordthompson , I haven't been able to properly test this though, so do please check what I've done.


mkleehammer:
Ha.  The new error message I added would not have helped in this case.  One of the checks below is for a sequence, which bytes are.  Thanks again.

PR #1300, opened 2023-11-01
AppVeyor Python 3.12 and other updates

keitherskine:
Python 3.12 is now available on AppVeyor Windows images.  Whilst we're adding that to the mix, I thought we might as well update the build we use too.  So this PR uses the Visual Studio 2022 Windows image, instead of 2019.  It also uses SQL Server 2019 instead of 2017, Postgresql 13 instead of 11, and MySQL8.0 instead of 5.7 which is now past its EOL.

You have to start MSSQL 2019 with a "net start" command, and MySQL 8.0 with a "Start-Service" [command](https://github.com/appveyor/ci/issues/3894#issuecomment-1787625858) but hopefully those will be cleaned up by AppVeyor at some point.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,20 +1,20 @@
 # This AppVeyor CI configuration file:
 #   - builds pyodbc with multiple versions of Python
 #   - tests the generated pyodbc module against various databases and drivers
-#   - creates "wheel" files for distribution, and stores them as appveyor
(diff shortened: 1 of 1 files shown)


mkleehammer:
Fantastic - thanks.  It's amazing that open source has free tools like this.  Thanks for keeping it up to date.

PR #1305, opened 2023-11-21
Populate cursor "messages" attribute when fast_executemany=True

### LINKED ISSUE #1299: Unable to obtain INFO messages using fast_executemany from SQL Server connection
- Python:  Python 3.11.6 (tags/v3.11.6:8b6ee5b, Oct  2 2023, 14:57:12) [MSC v.1935 64 bit (AMD64)] on win32
- pyodbc: 5.0.1
- OS:  Windows 11
- DB: SQL Server
- driver: MSODBCSQL17.dll

### Issue

Attempting to retrieve the message when using executemany.  Using a single execution, I am able to output the messages, but I have not determined how to invoke the SQLGetDiagRec when using executemany.

The sample procedure shows the INFO messages when I also do a raiserror but if the raiserror is not present I am unable to output the print message(s).

Repro

'''
use tempdb
go

drop table if exists t
go
drop procedure if exists spTest
go

create table t
(
	id int,
	id2 int
)
go

create procedure spTest
	@id int,
	@id2 int
as
begin
	insert into t values (@id, @id2)
	print 'In the procedure'
	-- raiserror('Bob', 16, 1)  -- If I add this then I get all the print messages
end
go

'''

import pyodbc

connection_string = ''' DRIVER={ODBC Driver 17 for SQL Server};
                        SERVER=localhost;
                        DATABASE=tempdb;
                        UID=sa;
                        PWD=xxxxxxxxxxx;
                    '''

# Creates the connection
conn = pyodbc.connect(connection_string)
conn.autocommit = False

cursor = conn.cursor()
params = [(i,i) for i in range(10)]
cursor.fast_executemany = True

cursor.executemany("{call spTest(?, ?)}", params)

# This works but the ExecuteMany I am unable to get the calls to SQLGetDiagRec for messages
# cursor.execute("{call spTest(1, 2)}")

print(cursor.rowcount)
print(cursor.messages)
    
while(cursor.nextset()):
    print(cursor.messages)

cursor.commit()    #  Does not show any information message here either but this can have them
cursor.close()

keitherskine:
When fast_executemany is True, function ExecuteMulti() is called instead of execute().  ExecuteMulti() was not calling GetDiagRecs() to populate the cursor "messages" attribute.  This now happens.  I had to export GetDiagRecs() from cursor.cpp for that purpose.

Unit test added.  Also beefed up an existing unit test for unicode messages.

Fixes #1299 .

### DIFF
--- a/src/cursor.cpp
+++ b/src/cursor.cpp
@@ -577,7 +577,7 @@ static bool PrepareResults(Cursor* cur, int cCols)
 }
 
 
-static int GetDiagRecs(Cursor* cur)
+int GetDiagRecs(Cursor* cur)
 {
     // Retrieves all diagnostic records from the cursor and assigns them to the "messages" attribute.
 
--- a/src/cursor.h
+++ b/src/cursor.h
@@ -156,6 +156,8 @@ struct Cursor
     PyObject* messages;
 };
 
+int GetDiagRecs(Cursor* cur);
+
 void Cursor_init();
 
 Cursor* Cursor_New(Connection* cnxn);
--- a/src/params.cpp
+++ b/src/params.cpp
@@ -1674,6 +1674,11 @@ bool ExecuteMulti(Cursor* cur, PyObject* pSql, PyObject* paramArrayObj)
             goto ErrorRet8;
         }
 
+        if (rc == SQL_SUCCESS_WITH_INFO)
+        {
+            GetDiagRecs(cur);
+        }
+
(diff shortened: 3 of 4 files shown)


mkleehammer:
This looks really nice.  Thanks for beefing up the test.

I've never been fond of having two code paths that are so separate, regular and fast_execute_many.  I'll be surprised if there aren't a lot of things like this.  When the change was first submitted, I wanted to break things apart and make them use the same base components - especially binding parameters.

PR #1315, opened 2024-01-08
set statement attributes

iloveitaly:
You can set connection-level attributes via `set_attr` but not cursor-level settings. Note that I haven't had time for tests here yet, but can put them together if adding this to the project makes sense.

- fix: set timeout before attempting connection timeout
- feat: adding set_attr to cursor

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -976,6 +976,9 @@ static int Connection_settimeout(PyObject* self, PyObject* value, void* closure)
         return -1;
     }
 
+    // this same value is used for the cursor timeout
+    cnxn->timeout = timeout;
+
     SQLRETURN ret;
     Py_BEGIN_ALLOW_THREADS
     ret = SQLSetConnectAttr(cnxn->hdbc, SQL_ATTR_CONNECTION_TIMEOUT, (SQLPOINTER)(uintptr_t)timeout, SQL_IS_UINTEGER);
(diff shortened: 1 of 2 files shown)


mkleehammer:
It does seem like a missing piece of the API - thanks.  Could you add at least one test for PostgreSQL, MySQL, or SQL Server?

I'm also curious about the impetus for this.  Were you looking for a statement timeout?

mkleehammer:
I really like this idea.

I really appreciate the note in the doc string that only integers are supported right now.  That will help a lot.

eramitmittal:
This could be useful for setting SQL_ATTR_ENABLE_AUTO_IPD statement attribute as well.

src/pyodbcmodule.cpp
```diff
@@ -1107,6 +1110,10 @@ static const ConstantDef aConstants[] = {
     MAKECONST(SQL_OJ_NOT_ORDERED),
     MAKECONST(SQL_OJ_INNER),
     MAKECONST(SQL_OJ_ALL_COMPARISON_OPS),
+
+    // Misc
+    MAKECONST(SQL_IS_UINTEGER),
+    MAKECONST(SQL_IS_UINTEGER),
```
keitherskine, line 1116: Duplicate lines.  Should one of these be SQL_IS_INTEGER perhaps?

PR #1316, opened 2024-01-10
feat: add [pyodbc] and timestamp to tracing output

iloveitaly:
makes debugging quite a bit easier.


mkleehammer:
Makes sense...  I should have done that originally.  Thanks.

iloveitaly:
@mkleehammer friendly bump here!

PR #1317, opened 2024-01-10
docs: add more TRACE calls for easier debugging

iloveitaly:
Found this useful when debugging a handling `SQLFreeStmt` command on my driver (netsuite).

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -154,6 +154,8 @@ static bool ApplyPreconnAttrs(HDBC hdbc, SQLINTEGER ikey, PyObject *value, char 
         return false;
     }
 
+    TRACE("ApplyPreconnAttrs.SQLSetConnectAttrW key=%d value=%p vallen=%d\n", ikey, ivalue);
+
     Py_BEGIN_ALLOW_THREADS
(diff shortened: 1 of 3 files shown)


src/connection.cpp
```diff
@@ -154,6 +154,8 @@ static bool ApplyPreconnAttrs(HDBC hdbc, SQLINTEGER ikey, PyObject *value, char
         return false;
     }
 
+    TRACE("ApplyPreconnAttrs.SQLSetConnectAttrW key=%d value=%p vallen=%d\n", ikey, ivalue);
```
mkleehammer, line 157: I haven't tested this locally, but it appears there are 3 placeholders and 2 values.

src/params.cpp
```diff
@@ -1310,6 +1310,8 @@ void FreeParameterData(Cursor* cur)
 {
     // Unbinds the parameters and frees the parameter buffer.
 
+    TRACE("FreeParameterData");
```
mkleehammer, line 1313: Needs newline?

mkleehammer:
Looks good, but I added 3 comments in the Files changed tab using the web UI.  Can you take a look at those?  Thanks.

Issue #1124, opened 2022-11-17
Cannot import `pyodbc` with latest version on Apple Silicon


nleroy917:
### Environment

To diagnose, we usually need to know the following, including version numbers.  On Windows, be
sure to specify 32-bit Python or 64-bit:

- Python: `3.10.6`
- pyodbc: `4.0.35`
- OS: macOS Ventura (13.0)
- DB: N/A
- driver: N/A

### Issue
**overview:**
I upgraded my `pyodbc` version to the latest on PyPi (v`4.0.35`) and noticed I was getting a new error that I had never dealt with before. Specifically, when importing `pyodbc` I receive the following:
```
(0x0002): symbol not found in flat namespace '_SQLAllocHandle'
```
An issue for this was raised and closed a while ago; it references this exact problem: #885

**expected behavior:**
`pyodbc` should import without issue

**steps to reproduce:**
1. Create new virtual environment and activate:
```console
python -m venv .venv
source .venv/bin/activate
```
2. Install latest `pyodbc`:
```console
pip install pyodbc
```
3. Import `pyodbc` and observe error:
```console
python
>>> import pyodbc
>>> (0x0002): symbol not found in flat namespace '_SQLAllocHandle'
```

keitherskine:
Many thanks for the feedback, @nleroy917 .  Just as an experiment, could I ask you to try setting the LDFLAGS and CPPFLAGS environment variables before installing pyodbc, per the [wiki](https://github.com/mkleehammer/pyodbc/wiki/Install#installing-on-macosx)?  Many thanks.

I appreciate pyodbc should "just install" but all feedback is useful.

svintuss:
Hi @nleroy917! I've just had the same issue the other day. I managed to fix it with a `--no-binary` option which forces pip to compile the module from source instead of installing from precompiled wheel. 

`pip install --no-binary :all: pyodbc`

nleroy917:
Ahh. Is this specifiable in a `requirements.txt` file?

svintuss:
I would also like to upvote the issue because ver. 4.0.34 works fine without compiling from source.
E. g. you may force pip to install specific version of pyodbc with:

`pip install pyodbc==4.0.34`

nleroy917:
Thanks for the help! I'll close this.

keitherskine:
[...]

PR #1323, opened 2024-02-04
Use M1 runner for macOS arm64 wheels

keitherskine:
Now that GitHub has made an [ARM64 runner](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/) available - hurrah! - let's use it to create the problematic macOS ARM64 wheels.

In the GitHub Action artifacts script, I have separated out the builds for each platform into discrete jobs.  They are all slightly different and it seems sensible to have a specific job for each platform.

We were getting a bunch of deprecation warnings about old versions of node, so I have updated the versions of actions/setup-python from v4 to v5, actions/upload-artifact from v3 to v4, and github/codeql-action from v2 to v3.  I have also taken the opportunity to pin various actions to specific minor versions so we know exactly what is being run.

Lastly, I have swapped around the Appveyor builds so that Python 3.12 is done first and Python 3.8 is done last.  On the free plan, Appveyor jobs are sequential (and slow), and typically it's the Python 3.12 build we are most interested in, so do it first.

### DIFF
--- a/.github/workflows/artifacts_build.yml
+++ b/.github/workflows/artifacts_build.yml
@@ -1,4 +1,7 @@
 # Build all the artifacts for a release (i.e. the source distribution file and the various wheels)
+# https://docs.github.com/en/actions/using-jobs/choosing-the-runner-for-a-job
+# https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
+# https://cibuildwheel.readthedocs.io/en/stable/options/#options-summary
 
 name: Build the release artifacts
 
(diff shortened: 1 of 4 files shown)


keitherskine:
You can see examples of the wheels that are generated here: https://github.com/keitherskine/pyodbc/actions/runs/7775109101
(they can't be built in the main mkleehammer pyodbc repo until this PR is merged into master)

This should hopefully fix #1124 .

keitherskine:
I have tested the arm64 wheel on an M1 Macbook Pro.  If anybody can test it on an M3 that would be appreciated.

PR #1339, opened 2024-03-29
Fix memory leak for strings passed to TVP

### LINKED ISSUE #1338: Memory leak when passing strings within table value parameter to stored proc
### Environment

- Python: 3.11.6 (tags/v3.11.6:8b6ee5b, Oct  2 2023, 14:57:12) [MSC v.1935 64 bit (AMD64)]
- pyodbc: 5.1.0
- OS: Windows 10
- DB: Windows Server 2019 Standard (10.0)
- driver: ODBC Driver 17 for SQL Server

### Issue

Memory leak when string values supplied to table valued parameter of stored proc. Same data supplied as float does not leak memory.

### Example code

Following code should reproduce the error.

```python
import psutil
import gc
import pyodbc

process = psutil.Process()

# Prepare database with do nothing stored proc
"""
CREATE TYPE Test_Table_Type AS TABLE
(
    RowId     bigint,
    MeasureId int,
    NodeId    int,
    Value     float  -- doesn't make any difference if this is varchar(256)
)
GO

CREATE PROCEDURE Test_Proc
@Facts Test_Table_Type READONLY
AS
BEGIN
  SELECT COUNT(*) FROM @Facts
END

GO
"""


def print_memory_usage():
    memory_usage = process.memory_info().rss  # should this be rss or vms
    print(f"{memory_usage / (1024 * 1024.0):0.4f} MiB")


def run_test(cnxn, call_proc, fn):
    stmt = "{CALL Test_Proc (?)}"
    cursor = cnxn.cursor()
    x = 3.00000001
    y = 2
    z = 4
    for i in range(5):
        x += 1.0
        y += 1
        z += 1
        # Make data with different strings for each row
        data = [(i, z * i, y * i, fn(x * i)) for i in range(1_000_000)]
        if call_proc:
            cursor.execute(stmt, (data,))
        del data
        gc.collect()
        print_memory_usage()


if __name__ == "__main__":
    url = 'DRIVER={ODBC Driver 17 for SQL Server};SERVER=MY_SERVER;DATABASE=MY_DATABASE;Trusted_Connection=yes;'
    connection = pyodbc.connect(url)
    print_memory_usage()
    print("\n# Make data but don't execute stored proc (with floats)")
    run_test(connection, call_proc=False, fn=float)
    print("\n# Make data but don't execute stored proc (with strings)")
    run_test(connection, call_proc=False, fn=str)
    print("\n# Make data and execute stored proc (with floats)")
    run_test(connection, call_proc=True, fn=float)
    print("\n# Make data and execute stored proc (with strings)")
    run_test(connection, call_proc=True, fn=str)
```

### Output of sample code

```
44.1406 MiB

# Make data but don't execute stored proc (with floats)
45.9414 MiB
47.8555 MiB
47.8633 MiB
47.8672 MiB
47.8672 MiB

# Make data but don't execute stored proc (with strings)
47.8750 MiB
47.8789 MiB
47.8789 MiB
46.1562 MiB
47.8242 MiB

# Make data and execute stored proc (with floats)
46.4648 MiB
48.1289 MiB
46.4727 MiB
48.1406 MiB
46.4766 MiB

# Make data and execute stored proc (with strings)
121.0195 MiB
194.1836 MiB
267.8984 MiB
344.1250 MiB
419.3477 MiB

Process finished with exit code 0
```

mavxg:
Fixes #1338

### DIFF
--- a/src/cursor.cpp
+++ b/src/cursor.cpp
@@ -873,8 +873,6 @@ static PyObject* execute(Cursor* cur, PyObject* pSql, PyObject* params, bool ski
                         Py_XDECREF(prevParam->pObject);
                         newParam.BufferLength = newParam.StrLen_or_Ind;
                         newParam.StrLen_or_Ind = SQL_DATA_AT_EXEC;
-                        Py_INCREF(cell);
-                        newParam.pObject = cell;
                         *prevParam = newParam;
                         if(prevParam->ParameterValuePtr == &newParam.Data)
                         {
--- a/src/params.cpp
+++ b/src/params.cpp
@@ -1266,6 +1266,7 @@ bool BindParameter(Cursor* cur, Py_ssize_t index, ParamInfo& info)
             {
                 // Bind the TVP's columns --- all need to use DAE
                 PyObject *param = PySequence_GetItem(row, i);
+                Py_XDECREF(param);
                 GetParameterInfo(cur, i, param, info.nested[i], true);
                 info.nested[i].BufferLength = info.nested[i].StrLen_or_Ind;
                 info.nested[i].StrLen_or_Ind = SQL_DATA_AT_EXEC;


gordthompson:
Looks okay to me, and it does appear to plug the leak. Thanks!

Should we include the Py_XDECREF call from [here](https://github.com/mkleehammer/pyodbc/issues/1338#issuecomment-2023133256) as well?

(My C++ skills are way too rusty to accept this PR myself.)

mavxg:
Yes. It fixes another leak for the first row/tuple in the sequence. For TVP, BindParameters is getting just the first row (not sure what the reason behind the `PySequence_Size(info.pObject) - info.ColumnSize` to get the first item in the sequence). `PyObject *param = PySequence_GetItem(row, i);` will increase the ref count of the items in the first tuple in the sequence and the Py_XDECREF would then decrease it again.

Issue #1344, opened 2024-04-16
PyToCType uses internal function removed from Python 3.13


mkleehammer:
Cannot compile using 3.13:

```
src/params.cpp: In function ‘int PyToCType(Cursor*, unsigned char**, PyObject*, ParamInfo*)’:
src/params.cpp:250:36: error: too few arguments to function ‘int _PyLong_AsByteArray(PyLongObject*, unsigned char*, size_t, int, int, int)’
  250 |             if (_PyLong_AsByteArray((PyLongObject*)absVal, pNum->val, sizeof(pNum->val), 1, 0))
      |                 ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

We need to figure out how to bind Decimals without using this, I guess.

v-chojas:
We're not the only ones affected by this regression: https://github.com/python/cpython/issues/111140
pyodbc is mentioned there along with some other projects.

AdamWill:
it seems a public function `PyLong_AsNativeBytes` has been added to do this job, IIUC. See https://github.com/python/cpython/blob/main/Doc/c-api/long.rst for docs on it.

As the error you get - "too few arguments" - implies, the function was re-added after it was removed, but then an extra argument, `with_exceptions` (an int acting as a boolean), was added in [the same commit that introduced `PyLong_AsNativeBytes`](https://github.com/python/cpython/commit/7861dfd26a). So for the short term, I think just adding the extra arg to the call would work. For the long term probably a good idea to switch to using the new public function, though.

AdamWill:
Sent https://github.com/mkleehammer/pyodbc/pull/1361 .

PR #1361, opened 2024-06-12
Adjust for _PyLong_AsByteArray signature change in Python 3.13 (#1344)

AdamWill:
PY_VERSION_HEX is documented at https://docs.python.org/3/c-api/apiabiversion.html , I think I got it right.


gordthompson (approved):
I have verified that this builds successfully under Python 3.13.0rc1 on Ubuntu 22.04 and can connect to SQL Server 2019 via msodbcsql18.

Issue #1438, opened 2025-08-19
[Debian 13] Segmentation fault on attempt to obtain a connection


vdmitrienko:
After [the release of Debian 13](https://www.debian.org/News/2025/20250809), we began to experience a segmentation fault on attempt to obtain a connection:
```python
    conn_params = {
        'DRIVER': driver,
        'ADDRESS': url,
        'autocommit': True,
    }
    conn = connect(**conn_params)
```
```
Fatal Python error: Segmentation fault
Current thread 0x0000ffffa58f7760 (most recent call first):
  File "/root/odbc_client.py", line 25 in _connect_odbc
  File "/root/odbc_client.py", line 52 in __execute_query
  File "/root/odbc_client.py", line 113 in main
  File "/root/odbc_client.py", line 117 in <module>
Extension modules: pyodbc (total: 1)
Segmentation fault
```
Debugger output:
```
# gdb python3
(gdb) run /root/odbc_client.py "Apache Ignite 3" "localhost:10800" "/root/queries.txt"
Starting program: /usr/bin/python3 /root/odbc_client.py "Apache Ignite 3" "localhost:10800" "/root/queries.txt"
warning: Error disabling address space randomization: Operation not permitted
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/aarch64-linux-gnu/libthread_db.so.1".
pyodbc version: 5.2.0
Program received signal SIGSEGV, Segmentation fault.
0x0000ffff83337ad0 in ?? () from /usr/lib/python3/dist-packages/pyodbc.cpython-313-aarch64-linux-gnu.so
```
The issue does not occur on Debian 12.

Environment:

- Debian version:
```
cat /etc/*-release
PRETTY_NAME="Debian GNU/Linux 13 (trixie)"
NAME="Debian GNU/Linux"
VERSION_ID="13"
VERSION="13 (trixie)"
VERSION_CODENAME=trixie
DEBIAN_VERSION_FULL=13.0
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
# uname -m
aarch64
```
- python version: `3.13.5`
- [pyodbc](https://pypi.org/project/pyodbc/5.2.0/) version: `5.2.0`

PR #1448, opened 2025-10-19
Enable assertions for test builds and run sqlite tests

shramov:
Follow up to #1440.

Enable assertions (pass -UNDEBUG) and run SQLite tests so bugs like #1438 will not slip in accidentialy.

Test for empty connection string is added in c2770d4bc35b3bef4f039ae1797bdf4f037c9d9b

### DIFF
--- a/.github/workflows/ubuntu_build.yml
+++ b/.github/workflows/ubuntu_build.yml
@@ -186,7 +186,7 @@ jobs:
         echo "*** current python version"
         python -VV
         echo "*** pip install pyodbc"
-        python -m pip install .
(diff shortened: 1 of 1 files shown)


mkleehammer:
Interesting - thanks.  I had not thought of this.  I'm going to hold off releasing a new minor / patch version for a few days to see what else turns up or I can fix.

shramov:
Sorry for spamming with pull requests but this one does not make sense if sqlite tests were not accepted...
If you are interested I have 'const char *' cleanup patch that fixes -Wno-write-strings warnings

PR #1452, opened 2025-12-15
Avoid heap corruption with NULLs in TVPs (#1450)

### LINKED ISSUE #1450: Executing stored procedures using TVP hard crashes python GC in python 3.11
Hello, 

Executing stored procedures using TVP hard crashes python GC in python 3.11. We think the issue has something to do with freeing memory inside `cursor.execute(sql, *parameters)`. Which then gets double freed inside `gc.collect()`.

**Environment**
- Python: 3.11 (It seems to also occur in 3.13)
- pyodbc: 5.3.0
- DB: MS SQL Server 2022
- Driver: Microsoft ODBC Driver 18 for SQL Server (18.5.1.1)

After writing using a stored procedure that uses a TVP to write data the python process crashes running `gc.collect()`. params contains regulair parameters and one set of TVP. 

Example:
```Python
conn = pyodbc.connect(connection_string, attrs_before={1256: bytearray(token_struct)})
sql = f'EXEC [{schema}].[{procedure}]'

# Params contains the TPV values as `{data : (Tuple)}`
with self._connection.cursor() as cursor:
    if params:
        sql += ' ' + ', '.join(f'@{p}=?' for p in params.keys())
        asdf = tuple(params.values())
        cursor.execute(sql, asdf)
    else:
        cursor.execute(sql)
  gc.collect(0)
  gc.collect(1)
  gc.collect(2)
```
CPython crashes inside `gc.collect(2)` without any form of exception. 

**Result**
```
Process finished with exit code -1073741819 (0xC0000005)
```
When GC is enabled, it crashes randomly. We think when garbage collection is triggered. With `gc.disable()` it doesn't crash. 

If you need any more information please let me know.

bkline:
The `GetParamType()` function calls `SQLDescribeParam()` and caches the type information it gets from the driver in a dynamically allocated array with one slot for each top-level parameter. This should not be done for nested table column "parameters" because:

- that corrupts the heap when a TVP has more columns than slots;
- it overwrites top-level parameter types with the wrong values;
- `SQLDescribeParam()` fails for nested column "parameters" anyway.

So for NULL TVP columns we fall back on SQL_VARCHAR.

With the fix contained in this commit, the new test never fails (for my testing, at least). Without the fix, that test sometimes passes and sometimes crashes the Python interpreter (and *always* crashed Python when the `tests/__pycache__` directory was not present in my testing--not sure why 🤷‍♂️).

Closes #1450

### DIFF
--- a/src/params.cpp
+++ b/src/params.cpp
@@ -534,9 +534,15 @@ static void FreeInfos(ParamInfo* a, Py_ssize_t count)
     PyMem_Free(a);
 }
 
-static bool GetNullInfo(Cursor* cur, Py_ssize_t index, ParamInfo& info)
+static bool GetNullInfo(Cursor* cur, Py_ssize_t index, ParamInfo& info, bool isTVP)
 {
-    if (!GetParamType(cur, index, info.ParameterType))
+    // GetParamType won't work for TVP columns, so we fall back on SQL_VARCHAR.
+    if (isTVP)
+    {
+        if (info.ParameterType == SQL_UNKNOWN_TYPE)
+            info.ParameterType = SQL_VARCHAR;
+    }
+    else if (!GetParamType(cur, index, info.ParameterType))
         return false;
 
     info.ValueType     = SQL_C_DEFAULT;
@@ -1023,7 +1029,7 @@ bool GetParameterInfo(Cursor* cur, Py_ssize_t index, PyObject* param, ParamInfo&
     // Populates `info`.
 
     if (param == Py_None)
-        return GetNullInfo(cur, index, info);
+        return GetNullInfo(cur, index, info, isTVP);
 
     if (param == null_binary)
         return GetNullBinaryInfo(cur, index, info);
--- a/tests/sqlserver_test.py
+++ b/tests/sqlserver_test.py
@@ -1,5 +1,6 @@
 #!/usr/bin/python
 
+import gc
 import os
 import re
 import uuid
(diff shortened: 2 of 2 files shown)


mkleehammer:
Thanks @bkline — your root-cause writeup made this straightforward to confirm, and I wanted something stronger than a lucky crash, so I reproduced it under AddressSanitizer (Python 3.12; msodbcsql 18.06.0002, so the unrelated #1459 driver regression doesn't interfere). The repro is a standalone version of your `test_tvp_with_nulls`: a 100-column TVP whose second row is all `NULL`, against a proc with a single top-level parameter.

On `master`, ASan flags it deterministically:

```
ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 2 ... 0 bytes after 2-byte region
  #0 GetParamType       src/params.cpp:1836
  #1 GetNullInfo        src/params.cpp:539
  #2 GetParameterInfo   src/params.cpp:1026
  #3 execute            src/cursor.cpp:856
allocated by thread T0 here:
  #0 malloc
  #1 GetParamType       src/params.cpp:1825
```

That 2-byte allocation is exactly `paramcount (1) * sizeof(SQLSMALLINT)` — the array sized for the single top-level parameter — and indexing it by TVP column position walks off the end, precisely as you described. With 100 columns it runs a long way past; this is the heap corruption that surfaced as the random `gc.collect()` crashes.

With the patch, the same repro is clean under ASan (no report, exit 0), and `test_tvp_with_nulls` plus the rest of the TVP/param/decimal tests pass under ASan with no findings.

|                          | master                                | this PR |
|--------------------------|---------------------------------------|---------|
| repro under ASan         | heap-buffer-overflow in `GetParamType`| clean   |
| `test_tvp_with_nulls`    | (added here)                          | passes  |

The `SQL_VARCHAR` fallback for TVP columns is the right call — same approach we already use when the driver can't describe a parameter. Merging.

PR #1460, opened 2026-02-11
Fix .statistics() issue with Unicode table names

### LINKED ISSUE #1457: Cursor.Statistics does not return values for an Access table with Kanji characters
I have an access database that I am trying to get index information for.  When my table is named with Japanese/Kanji characters, the statistic function fails to return the information.

# Repo:
- Extract the zip [kanji_pyodbc.zip](https://github.com/user-attachments/files/24564680/kanji_pyodbc.zip)
- Run stats_test_kanji.py(code is below too)
- Review results on MyTable vs ランドマーク

Results from script
```
MyTable
[('C:\\temp\\kanji_pyodbc\\kanji.mdb', None, 'MyTable', None, 'MyTable', None, 0, None, None, None, 0, 0, None), ('C:\\temp\\kanji_pyodbc\\kanji.mdb', None, 'MyTable', 0, 'MyTable', 'PrimaryKey', 3, 1, 'ID', 'A', 0, 1, None), ('C:\\temp\\kanji_pyodbc\\kanji.mdb', None, 'MyTable', 1, 'MyTable', 'foo', 3, 1, 'foo', 'A', 0, 1, None)]
ランドマーク
[]
```

script
```py
import pyodbc
import pathlib

DRIVER = "Microsoft Access Driver (*.mdb, *.accdb)"

path = pathlib.Path(__file__).parent.absolute()
mdb = path / 'kanji.mdb'
conn = pyodbc.connect(f"DRIVER={{{DRIVER}}};DBQ={str(mdb)}")
conn.setdecoding(pyodbc.SQL_CHAR, encoding='utf-16-le')
conn.setdecoding(pyodbc.SQL_WCHAR, encoding='utf-16-le')
conn.setdecoding(pyodbc.SQL_WMETADATA, encoding='utf-16-le')
conn.setencoding(encoding='utf-16-le')

with conn.cursor() as cursor:
    tables = [t.table_name for t in cursor.tables(tableType="TABLE")]
cursor.close()
del cursor

print(tables)

with conn.cursor() as cursor:
    for table_name in tables:
        print(table_name)
        print(cursor.statistics(table=table_name).fetchall())
cursor.close()
del cursor
```

gordthompson:
Fixes: #1457

Patch courtesy of @bkline

### DIFF
--- a/src/cursor.cpp
+++ b/src/cursor.cpp
@@ -1521,17 +1521,19 @@ char* Cursor_statistics_kwnames[] = { "table", "catalog", "schema", "unique", "q
 
 static PyObject* Cursor_statistics(PyObject* self, PyObject* args, PyObject* kwargs)
 {
-    const char* szCatalog = 0;
-    const char* szSchema  = 0;
-    const char* szTable   = 0;
+    PyObject* pCatalog = 0;
+    PyObject* pSchema  = 0;
+    PyObject* pTable   = 0;
     PyObject* pUnique = Py_False;
     PyObject* pQuick  = Py_True;
 
-    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "s|zzOO", Cursor_statistics_kwnames, &szTable, &szCatalog, &szSchema,
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O|OOOO", Cursor_statistics_kwnames, &pTable, &pCatalog, &pSchema,
                                      &pUnique, &pQuick))
         return 0;
 
     Cursor* cur = Cursor_Validate(self, CURSOR_REQUIRE_OPEN);
+    if (!cur)
+        return 0;
 
     if (!free_results(cur, FREE_STATEMENT | FREE_PREPARED))
         return 0;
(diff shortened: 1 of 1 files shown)


mkleehammer:
Confirmed and reproduced. On master, `statistics()` returns no rows for a table whose name is non-ASCII -- it hands the UTF-8 bytes to the ANSI `SQLStatistics`, so the driver matches nothing -- and with this patch it returns the index rows, with ASCII names unaffected. Encoding the arguments with the connection's `TextEnc` and calling `SQLStatisticsW` is the right fix; it brings `statistics()` in line with the other catalog functions. The change still applies cleanly to current master, and the full suite passes with it here (SQLite, PostgreSQL, SQL Server).

On the test @bkline found hard to write: the reason is connection pooling. The mis-encoded lookup fails on a fresh physical connection, but a reused pooled connection comes back warmed and the lookup succeeds, so the bug is masked -- which is why it only showed up as the first test in a run and vanished otherwise. Forcing a fresh connection makes it deterministic: a unique `APP=` keeps pooling from handing back a warmed connection, and then it reproduces every time, even when the test runs last. I pushed such a test to the branch (`test_statistics_unicode`); it fails without the cursor.cpp change and passes with it.

Thanks @gordthompson.

PR #1463, opened 2026-02-26
Add support for SQLTablePrivileges

### LINKED ISSUE #1340: Add support for SQLTablePrivileges
### Environment

- Python: 3.11
- pyodbc: 5.1.0
- OS: Linux
- DB: Db2
- driver: libdb2o.so

### Issue

Add support for the ODBC function [SQLTablePrivileges](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqltableprivileges-function?view=sql-server-ver16)

bkline:
We're limited in what we can test, as we don't know that we'll be running with GRANT permissions in the pipeline test harness, but we can at least make sure the function returns a results set with the right shape and column names. No point in testing on SQLite, which doesn't implement a PRIVILEGES model.

Separate PR for Wiki docs to follow.

I also suppressed .DS_Store in .gitignore.

Closes #1340

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,7 @@ pyodbc.conf
 tmp
 tags
 xxx_*
+.DS_Store
 
 # The Access unit tests copy empty.accdb and empty.mdb to these names and use them.
 test.accdb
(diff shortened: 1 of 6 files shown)


bkline:
Don't see a button for forking the wiki, and I don't have permission to push a branch to that repo, so here's a patch for the documentation.

[1340-wiki.diff.txt](https://github.com/user-attachments/files/25588794/1340-wiki.diff.txt)

(Pardon the strange filename ending. GitHub wouldn't accept a `.diff` file.)

PR #1465, opened 2026-03-04
Allow setting of connection string attributes

### LINKED ISSUE #505: Add support for changing the current catalog via Connection#set_attr
The Python DB API makes no mention of changing the current catalog. However, `System.Data.Odbc` in .NET has a `Connection#ChangeDatabase` method and

```csharp
cnxn.ChangeDatabase("master")
```

results in

```
OdbcConsoleApp. 2414-2578   EXIT  SQLSetConnectAttrW  with return code 1 (SQL_SUCCESS_WITH_INFO)
        SQLHDBC             0x0623EB90
        SQLINTEGER                 109 <SQL_ATTR_CURRENT_CATALOG>
        SQLPOINTER          0x01FE57FC [      12] "master"
        SQLINTEGER                  12

        DIAG [01000] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Changed database context to 'master'. (5701)
```

so in Python I tried

```python
cnxn.set_attr(pyodbc.SQL_ATTR_CURRENT_CATALOG, "master")
```

but I got

```
  File "C:/Users/Gord/PycharmProjects/py3pyodbc_demo/main.py", line 96, in <module>
    cnxn.set_attr(pyodbc.SQL_ATTR_CURRENT_CATALOG, "master")
TypeError: an integer is required (got type str)
```

If `set_attr` could be modified to also accept a string value then we'd have a database-independent way of changing the current catalog.

bkline:
I don't see any evidence that the regression tests are run with the ability to create a database, so I can't write a test that verifies the behavior requested by the OP (though I have tested locally to confirm that it works as expected with a DBMS which supports switching the current database for an existing connection; see ticket).

Fixes #505

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -351,25 +351,44 @@ static char set_attr_doc[] =
     "attr_id\n"
     "  The attribute id (integer) to set.  These are ODBC or driver constants.\n\n"
     "value\n"
-    "  An integer value.\n\n"
-    "At this time, only integer values are supported and are always passed as SQLUINTEGER.";
+    "  An integer or string value.";
 
 static PyObject* Connection_set_attr(PyObject* self, PyObject* args)
 {
     int id;
-    int value;
-    if (!PyArg_ParseTuple(args, "ii", &id, &value))
+    PyObject* value;
+    if (!PyArg_ParseTuple(args, "iO", &id, &value))
         return 0;
 
     Connection* cnxn = (Connection*)self;
 
     SQLRETURN ret;
-    Py_BEGIN_ALLOW_THREADS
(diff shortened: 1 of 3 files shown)


bkline:
Hold off on review until I add a test.

bkline:
OK, the test has been added. Had a moment of mild panic, before I realized that the failing test wasn't related to this PR. Perhaps related to #1459 (though the test which was seg-faulting in this case was `test_cursor_with_print()`, not `test_tvp()` as reported in that ticket)? I did confirm that `pyodbc` dumps core with `test_cursort_with_print()` (I haven't yet downgraded my 18.6.1.1 SQL Server driver) without this PR, so I'm reasonably confident that it's not related to this patch.

PR #1472, opened 2026-03-30
Keep SetConnectAttrW buffers alive

### LINKED ISSUE #1469: Use-after-free in `pyodbc.connect()`: `attrs_before` `bytes` buffer passed to GIL-released `SQLSetConnectAttrW()` without a strong reference
## Environment

- Python: 3.13.2 / 3.14.2
- pyodbc: 5.3.0
- OS: Linux x86_64, Windows x64
- DB/driver: Azure SQL Database, ODBC Driver 18 for SQL Server

## Summary

`pyodbc.connect(..., attrs_before=...)` has a lifetime bug when a top-level `attrs_before` value is `bytes`.

In `src/connection.cpp` (pyodbc 5.3.0), `ApplyPreconnAttrs()` takes a raw pointer from `PyBytes_AsString(value)` at lines 127-131, then releases the GIL at lines 157-159 before calling `SQLSetConnectAttrW()`:

```cpp
else if (PyBytes_Check(value))
{
    ivalue = PyBytes_AsString(value);
    vallen = SQL_IS_POINTER;
}

Py_BEGIN_ALLOW_THREADS
ret = SQLSetConnectAttrW(hdbc, ikey, ivalue, vallen);
Py_END_ALLOW_THREADS
```

That `value` comes from `PyDict_Next()` in `Connection_New()`, so it is only a borrowed reference:

```cpp
while (PyDict_Next(attrs_before, &pos, &key, &value))
{
    ...
    if (!ApplyPreconnAttrs(hdbc, ikey, value, strencoding))
        return 0;
}
```

The dict itself is `Py_INCREF`'d in `_CheckAttrsDict()`, but the individual `bytes` object is not.

If another thread replaces `attrs_before[key]` during the GIL-released window, the old `bytes` object can be decref'd to zero and freed while the ODBC driver is still reading from its buffer.

## Why the dict `INCREF` is not sufficient

Keeping the dict alive does not keep the old value alive.

If thread A is inside `SQLSetConnectAttrW()` using a pointer obtained from `attrs_before[key]`, and thread B does:

```python
attrs_before[key] = new_bytes
```

then the old `bytes` value is decref'd. If that was its last reference, its memory is freed even though thread A is still using the pointer derived from it.

## Observed result

In production this showed up as:

- Linux: intermittent SIGSEGV
- Windows: access violations / corrupted driver state

This originally surfaced through concurrent `pyodbc.connect()` calls using a shared mutable `attrs_before` dict carrying an Azure access token.

## Suggested fix direction

Before releasing the GIL, pyodbc should keep a strong reference to the exact `bytes` object whose buffer is passed to `SQLSetConnectAttrW()`, and release that reference after the call returns.

Something like:

```cpp
PyObject* keepalive = 0;

if (PyBytes_Check(value))
{
    keepalive = value;
    Py_INCREF(keepalive);
    ivalue = PyBytes_AsString(value);
    vallen = SQL_IS_POINTER;
}

Py_BEGIN_ALLOW_THREADS
ret = SQLSetConnectAttrW(hdbc, ikey, ivalue, vallen);
Py_END_ALLOW_THREADS

Py_XDECREF(keepalive);
```

I have only minimized the `bytes` case here. `bytearray` may need different handling because keeping the object alive does not guarantee its internal buffer stays stable across mutation.

## Related context

This was originally tracked downstream in SQLAlchemy when shared mutable connect-hook state was reused across concurrent connects:

- https://github.com/sqlalchemy/sqlalchemy/issues/13144

SQLAlchemy added a caller-side workaround by copying per-call collections, but the underlying pyodbc lifetime issue remains for callers that share a mutable `attrs_before` dict across threads.

bkline:
According to the ODBC specification, only a small handful of API functions are allowed to keep pointers for buffers of non-scalar values for use after the function has returned. The spec identifies each of those functions, and SetConnectAttrW is not one of them. However, Microsoft's ODBC Driver 18 for SQL Server does not comply with this restriction, and if we don't keep such buffers valid until after SQLDriverConnectW has returned, the application can crash. See https://github.com/microsoft/msphpsql/issues/1594 for further details.

Closes #1469

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -99,13 +99,11 @@ static bool Connect(PyObject* pConnectString, HDBC hdbc, long timeout, PyObject*
     return false;
 }
 
-static bool ApplyPreconnAttrs(HDBC hdbc, SQLINTEGER ikey, PyObject *value, char *strencoding)
+static bool ApplyPreconnAttrs(HDBC hdbc, SQLINTEGER ikey, PyObject *value, char *strencoding, PyObject *keepalives)
 {
     SQLRETURN ret;
     SQLPOINTER ivalue = 0;
     SQLINTEGER vallen = 0;
-
-    SQLWChar sqlchar;
 
     if (PyLong_Check(value))
     {
@@ -121,19 +119,33 @@ static bool ApplyPreconnAttrs(HDBC hdbc, SQLINTEGER ikey, PyObject *value, char 
     }
     else if (PyByteArray_Check(value))
     {
+        // Keep the value alive beyond this function invocation's lifetime.
+        if (PyList_Append(keepalives, value))  // OOM
+            return false;
         ivalue = (SQLPOINTER)PyByteArray_AsString(value);
         vallen = SQL_IS_POINTER;
     }
     else if (PyBytes_Check(value))
     {
-        ivalue = PyBytes_AsString(value);
+        // Keep the value alive beyond this function invocation's lifetime.
+        if (PyList_Append(keepalives, value))  // OOM
+            return false;
+        ivalue = (SQLPOINTER)PyBytes_AsString(value);
         vallen = SQL_IS_POINTER;
     }
     else if (PyUnicode_Check(value))
     {
-        sqlchar.set(value, strencoding ? strencoding : "utf-16le");
-        ivalue = sqlchar.get();
+        PyObject* po = PyUnicode_AsEncodedString(value, strencoding ? strencoding : "utf-16le", "strict");
+        if (!po)
+            return false;  // OOM
+        ivalue = (SQLPOINTER)PyBytes_AsString(po);
         vallen = SQL_NTS;
+        // Keep the value alive beyond this function invocation's lifetime ...
+        int failed = PyList_Append(keepalives, po);
+        // ... but not forever.
+        Py_DECREF(po);
+        if (failed)  // OOM
+            return false;
     }
     else if (PySequence_Check(value))
     {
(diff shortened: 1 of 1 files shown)


mkleehammer:
Thanks for the very thorough investigation here, @bkline — especially standing up an Azure account to get a real repro and tracing it to the driver's [deferred use of the buffers](https://github.com/microsoft/msphpsql/issues/1594). The diagnosis matches what I see in the code: `ApplyPreconnAttrs` was handing borrowed buffer pointers to the driver — the `bytes`/`bytearray` storage and, for `str`, the stack-allocated `SQLWChar` that's destroyed the moment the function returns — none of which are guaranteed to outlive the call, let alone survive until `SQLDriverConnectW`, which is what the MS driver actually requires.

I verified the change locally before merging:

- Full SQLite + PostgreSQL suites on Python 3.10 / 3.12 / 3.14 — no regressions (113 passing, same as master).
- A focused refcount check that connects repeatedly with the *same* `bytes` / `bytearray` / `str` object in `attrs_before`: the new keepalive list leaves each object's refcount unchanged (delta 0) on both the success and error paths over 2000 iterations — no leak, no over-release.

One note as a possible follow-up (not a blocker): **`bytearray` is only partially covered.** Appending it to `keepalives` keeps the *object* alive, but `PyByteArray_AsString` returns a pointer into its internal storage, which can be reallocated if the buffer is resized (e.g. another thread does `attrs[key].extend(...)`). For the deferred-read driver bug this targets that's fine — nothing mutates it between the set and the connect, and the immutable `bytes`/`str` paths are fully safe — but the *original* concurrent-mutation scenario from the issue isn't fully closed for `bytearray`. We could snapshot it into `bytes` for the keepalive, or just document that `bytearray` attr values mustn't be mutated concurrently during connect.

Going to merge this. Thanks again!

PR #1473, opened 2026-04-01
Prevent tests from failing with FreeTDS

### LINKED ISSUE #1471: Tests fail when using the FreeTDS driver
## Summary

Some newer tests in `sqlserver_test.py` fail when run against the FreeTDS ODBC driver, and one test (`test_cursor_messages_with_fast_executemany`) actually [crashes the test suite ](https://github.com/FreeTDS/freetds/issues/716)when running with a version of FreeTDS older than version 1.5.15. Other tests which do not pass:

```log
=========================================== short test summary info ============================================
FAILED tests/sqlserver_test.py::test_varbinary - TypeError: object of type 'NoneType' has no len()
FAILED tests/sqlserver_test.py::test_rowcount_delete - assert 0 == -1
ERROR tests/sqlserver_test.py::test_no_fetch - pyodbc.ProgrammingError: ('25000', '[25000] [FreeTDS][SQL Server]The COMMIT TRANSACTION request has no corr...
=============================== 2 failed, 90 passed, 5 skipped, 1 error in 3.13s ===============================
```

## Expected

Tests known to fail with a given ODBC driver are skipped when that driver is detected.

## Observed

Three tests do not pass when run with FreeTDS and one test crashes the test suite. Also, there is a typo in the string used to detect FreeTDS ("tsodbc" should be "tdsodbc").

## Resolution

I will submit a PR which addresses each of the failures. The solution for each test can involve one or more of the following:
- work with the FreeTDS team to get driver bugs fixed
- mark the test to be skipped when FreeTDS is detected
- mark the test to be skipped when FreeTDS older than a certain version is detected
- tweak the test to match the behavior of the driver

I will also fix the driver detection typo (in a way which does not break any test environments which rely on that typo).

I have been submitting tickets to the [FreeTDS project](https://github.com/FreeTDS/freetds) to get the issues addressed. That team has been very responsive, and a fix for one of the bugs has already been [released](https://github.com/FreeTDS/freetds/releases/tag/v1.5.15).

bkline:
Some tests have been tweaked, others skipped.

Also fixes a typo in string used to detect the FreeTDS drier.

Closes #1471

### DIFF
--- a/tests/sqlserver_test.py
+++ b/tests/sqlserver_test.py
@@ -1454,7 +1454,7 @@ def test_columns(cursor: pyodbc.Cursor):
         cursor.execute(f"drop table {table_name}")
 
 
-@pytest.mark.skipif(IS_FREEDTS, reason="FreeTDS Unicode handling for catalog functions is unreliable")
+@pytest.mark.skipif(IS_FREETDS, reason="FreeTDS Unicode handling for catalog functions is unreliable")
 def test_statistics_unicode():
     # https://github.com/mkleehammer/pyodbc/issues/1457
     # statistics() passed the table name straight to the ANSI SQLStatistics, mis-encoding a
@@ -1685,7 +1685,7 @@ def _test_tvp_with_nulls_cleanup(cursor: pyodbc.Cursor, procname: str, typename:
 
 
 @pytest.mark.skipif(SQLSERVER_YEAR < 2008, reason="TVP not supported until 2008")
-@pytest.mark.skipif(IS_FREEDTS, reason="FreeTDS does not support TVP")
+@pytest.mark.skipif(IS_FREETDS, reason="FreeTDS does not support TVP")
 def test_tvp_with_nulls(cursor: pyodbc.Cursor):
     """Make sure NULL values in a TVP don't crash the interpreter."""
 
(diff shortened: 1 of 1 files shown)


mkleehammer:
I verified this under FreeTDS locally -- version 1.3.17 (what Ubuntu 24.04 ships), against SQL Server 2022. On master the detection typo means none of the FreeTDS skips fire, and the suite segfaults at `test_cursor_messages_with_fast_executemany`. With this PR the detection is fixed (`tdsodbc` now matches the `libtdsodbc.so` driver name), the version gates all fire, and the crash is gone -- the suite goes from a hard crash to 91 passed, 7 skipped.

Two notes:

1. The rename needs a rebase. `IS_FREEDTS` became `IS_FREETDS`, but since this PR was opened master picked up another `IS_FREEDTS` use (in `test_tvp_with_nulls`, from #1452), and #1460 is about to add one more (`test_statistics_unicode`). Those aren't in this diff, so as-is the merge leaves a dangling `IS_FREEDTS` and a `NameError` on import. I'll rebase and finish the rename across all of them.

2. On 1.3.17 one failure remains -- `test_context_manager_failure` -- but it passes in isolation, so it's a cascade from the FreeTDS "COMMIT with no BEGIN" bug (#718) landing on a different test than the `conn2` workaround anticipated. The version gates here target 1.5.x, so 1.3.17 is likely just below the floor you validated against -- flagging it in case you want to widen the workaround or document a minimum FreeTDS version.

Thanks @bkline -- big improvement.

PR #1484, opened 2026-04-14
Implement PEP 675

### LINKED ISSUE #1166: Implement pep675
Python now provides LiteralString to help developers prevent security flaws from improperly handling user inputs. This would help puisne users.

bkline:
I hesitated with this one, after reading the [discussion in a similar ticket](https://github.com/sqlalchemy/sqlalchemy/issues/10255) for the SQLAlchemy project. But then I saw that @mkleehammer has tagged this for release with the [6.0 Stable ABI](https://github.com/mkleehammer/pyodbc/milestone/4) milestone, so here's the PR.

I took a fairly conservative approach, modifying the obvious SQL execution methods where the bulk of the vulnerabilities tend to bite. Some of the (less-frequently used) catalog methods could in theory also be vulnerable if a driver were to try and collect the requested information using hand-rolled SQL strings instead of parameterized queries (or the builtin stored procedures most back ends provide for this purpose), but that's much further into edge-case territory, and I didn't want to go there without some encouragement from the core team to do so (and I thought it best to keep this simple in case it's decided we don't want to support the PEP after all).

As a side note, if you want to test, you should be aware that [mypy](https://www.mypy-lang.org/) doesn't yet know what to do with `LiteralString`. You'll want to use the newer [pyright](https://github.com/microsoft/pyright) instead.

Closes #1166

### DIFF
--- a/src/pyodbc.pyi
+++ b/src/pyodbc.pyi
@@ -3,6 +3,7 @@
 from __future__ import annotations
 from collections.abc import Generator, Iterable, Iterator, Sequence
 from typing import Any, Callable, Final, Union
+from typing_extensions import LiteralString
 
 
 # SQLSetConnectAttr attributes
@@ -484,7 +485,7 @@ class Connection:
         """
         ...
 
-    def execute(self, sql: str, *params: Any) -> Cursor:
+    def execute(self, sql: LiteralString, *params: Any) -> Cursor:
         """A convenience function for running queries directly from a Connection object.
         Creates a new cursor, runs the SQL query, and returns the new cursor.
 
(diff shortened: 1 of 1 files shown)


mkleehammer:
Thanks for this, and for thinking about injection safety at the type level.

I've decided not to take this one. Typing the `sql` parameter as `LiteralString` makes a type checker reject any dynamically built query — an f-string, a concatenation, or just a query held in a variable — even when the construction is perfectly safe: interpolating a column or table name that never came from user input, assembling a statement from trusted fragments, building SQL in a helper and passing it in, and so on. Those patterns are common and legitimate in a lot of pyodbc code, so this would produce a wave of false positives that users have to silence with `cast()` or `# type: ignore`. It also pulls `typing_extensions` into the stub as a type-check dependency.

For a general-purpose DB API driver I'd rather keep `sql: str` and continue making the parameterize-don't-interpolate case in the docs, where it doesn't penalize correct dynamic SQL.

Closing this one, but I appreciate it -- and please don't let it discourage the rest of the PRs you've sent.

bkline:
Understood. Will you be closing the ticket as well (and removing the milestone)?

PR #1495, opened 2026-06-02
Add an optional SQL Server BCP fast path for fast_executemany

rad-pat:
## Summary

This adds an **optional SQL Server BCP (bulk copy) fast path** for
`Cursor.executemany()` when `fast_executemany` is enabled. When turned on, an
`INSERT ... VALUES (...)` executemany is routed through the SQL Server bulk-copy
API (`bcp_init` / `bcp_bind` / `bcp_sendrow` / `bcp_batch` / `bcp_done`) instead
of parameter-array `INSERT`s. For large loads this is dramatically faster than
`fast_executemany` alone, because it bypasses per-statement query processing.

It is **opt-in and safe**: if BCP can't be used for a given call (not a parseable
`INSERT`, an unsupported value type, a non-Microsoft driver, etc.) it
transparently falls back to the existing `fast_executemany` path.

The BCP entry points are **loaded on demand from the in-use `msodbcsql` driver**,
so there is no new build/link dependency.

## How to use (raw pyodbc)

BCP must be enabled on the connection *before* connecting, via the
SQL Server-specific `SQL_COPT_SS_BCP` pre-connect attribute:

```python
import pyodbc

SQL_COPT_SS_BCP = 1219
SQL_BCP_ON = 1

cnxn = pyodbc.connect(connection_string,
                      autocommit=True,
                      attrs_before={SQL_COPT_SS_BCP: SQL_BCP_ON})

cur = cnxn.cursor()
cur.fast_executemany = True      # required
cur.use_bcp_fast = True          # opt into the BCP path
cur.bcp_batch_rows = 10_000      # optional: rows per committed batch (0 = one batch)
cur.bcp_tablock = True           # optional: pass the TABLOCK hint

cur.executemany(
    "INSERT INTO dbo.target (a, b, c) VALUES (?, ?, ?)",
    rows)
```

New `Cursor` attributes:

| attribute | type | meaning |
|-----------|------|---------|
| `use_bcp_fast` | bool | route `fast_executemany` through BCP when possible |
| `bcp_batch_rows` | int | rows per committed batch (`0` = a single batch) |
| `bcp_tablock` | bool | pass the `TABLOCK` hint (bulk-update lock; can enable minimal logging) |

`AUTOCOMMIT` is recommended for BCP loads (the bulk copy is committed by
`bcp_done`/`bcp_batch`).

## How to use via SQLAlchemy

Enable BCP on the connection through `connect_args`, and flip `use_bcp_fast` on
each `executemany` with a `before_cursor_execute` listener:

```python
import sqlalchemy as sa
from sqlalchemy import event

SQL_COPT_SS_BCP = 1219
SQL_BCP_ON = 1

engine = sa.create_engine(
    "mssql+pyodbc://user:pwd@host/db?driver=ODBC+Driver+18+for+SQL+Server",
    connect_args={"attrs_before": {SQL_COPT_SS_BCP: SQL_BCP_ON}},
    fast_executemany=True,           # SQLAlchemy sets cursor.fast_executemany
    isolation_level="AUTOCOMMIT",
)

@event.listens_for(engine, "before_cursor_execute")
def _enable_bcp(conn, cursor, statement, parameters, context, executemany):
    if executemany:
        cursor.use_bcp_fast = True
        cursor.bcp_batch_rows = 10_000
        # cursor.bcp_tablock = True

with engine.begin() as conn:
    conn.execute(sa.insert(my_table), list_of_dicts)
```

## Supported value types

The host column types are deduced from the **first row**:

- `int` → `INT` (or `BIGINT` when it doesn't fit in 32 bits)
- `bool` → `BIT`
- `float` → `FLOAT`
- `str` / `bytes` / `bytearray` → character (varchar/nvarchar/...)
- `Decimal` → sent as text
- `datetime.date` / `datetime.time` / `datetime.datetime` (naive and tz-aware) → sent as text (ISO), letting the server convert to `DATE`/`TIME`/`DATETIME2`/`DATETIMEOFFSET`
- `None` → SQL `NULL`

Anything else (e.g. `uuid.UUID`) causes that `executemany` to fall back to the
standard `fast_executemany` path.

## Behaviour / fallback

`executemany` falls back to the existing `fast_executemany` implementation
(no error) when any of the following holds: the statement isn't a parseable
`INSERT`, the parameter sequence isn't a sequence of tuples, a column's value
type isn't supported, or BCP can't be loaded from the driver. So enabling
`use_bcp_fast` never breaks a statement that BCP can't handle.

## What is **not** done / limitations

- **SQL Server + `msodbcsql` only.** BCP is a SQL Server driver feature; other
  drivers/databases simply fall back.
- **Native binary temporal types are not enabled.** `date`/`time`/`datetime2`/
  `datetimeoffset` are sent as ISO text and converted server-side. The native
  length-prefixed binary bind is implemented but disabled: `bcp_bind` does not
  carry the fractional-second scale, so the driver rejects sub-second values
  (`22008 Fractional second precision exceeds the scale specified`). Enabling it
  needs the destination column's scale conveyed to BCP. `Decimal` is likewise
  sent as text.
- **Types are deduced from the first row.** If a later row needs a wider host
  type than the first row implied (e.g. an `int` > 32 bits after a first-row
  `int` that fit in 32 bits, or a non-`None` value in a column whose first value
  was `None`), it may be mis-bound. Provide representative first rows, or leave
  `use_bcp_fast` off for heterogeneous data.
- `bytes` are bound as character data, not `VARBINARY`.
- Reading a `DATETIMEOFFSET` column back is unrelated to this change but worth
  noting: pyodbc has no native decoder for SQL type `-155`, so round-trip tests
  read it via `CAST(... AS datetime2)` + `DATEPART(tzoffset, ...)`.

## Performance

Measured in CI inserting into a heap table with the four-column schema
`(INT, FLOAT, VARCHAR(100), DATETIME2(7))`. The first method uses a smaller row
count because the row-by-row path is slow; the bulk methods use 100k rows.
Database in **FULL** recovery:

| method | rows | time | rate |
|---|---:|---:|---:|
| `executemany` (row-by-row) | 5,000 | 3.36 s | ~1,500 rows/s |
| `fast_executemany` | 100,000 | 37.3 s | ~2,700 rows/s |
| BCP, no `TABLOCK` | 100,000 | 0.367 s | ~272,000 rows/s |
| BCP, `TABLOCK` | 100,000 | 0.370 s | ~270,000 rows/s |

A separate run under **SIMPLE** recovery produced effectively the same figures
(BCP ≈ 0.37–0.39 s either way).

Takeaways:

- BCP is roughly **two orders of magnitude** faster than `fast_executemany` here
  (~0.37 s vs ~37 s for 100k rows).
- Almost all of the gain comes from the **BCP protocol itself**, not from
  minimal logging: `TABLOCK` made a negligible difference at this size, and the
  result held under both FULL and SIMPLE recovery.
- **Caveat:** this is a small, shared CI runner where `fast_executemany` is
  abnormally slow (~2,700 rows/s), which inflates the multiple. On capable
  hardware the gap will be smaller, but BCP remains clearly faster for bulk
  loads.

## Tests

`tests/sqlserver_test.py::test_bcp_fast_executemany` exercises the path across
all supported types, `NULL`s, variable-length growth, `>32/64`-bit integers, and
manual batching, asserting exact round-trips (skipped on non-Microsoft drivers).
`tests/bcp_benchmark.py` is a standalone benchmark (not collected by pytest).
The full CI matrix (Python 3.9–3.14 × SQL Server 2019/2022/2025 + PostgreSQL /
MySQL / SQLite) is green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rad-pat:
Closing for now, a couple more issues to iron out.

